PR #2, opened 2013-06-18
Add support for passing main method arguments

celkins:


johnmcteague:
I am revisiting this topic; as many of our app developers look to migrate apps to CF, the issue of cli args rears its ugly head again.

Specifying a custom start command for java buildpacks to get CLI args is not the greatest user experience imo. As far as I can tell you loose the JAVA_HOME and JAVA_OPTS variables since the output of the release function is ignored when a custom command is specified, so users must know these and set them in the start command (if the location of the JDK ever changes within the container, all the custom start commands disappear). Any java_opts i want to enforce globally are lost.

There are many different application patterns, whether they be bespoke patterns employed by teams or third party frameworks like dropwizard. As I see a drive towards increased use of embedded servlet containers and wrapper frameworks, trying to add buildpack support becomes more and more cumbersome, especially as they all start to look very similar when trying to run detect over them but have very subtle differences in start command formats.

My experimental fork of the buildpack simply picks up JAVA_ARGS from the environment and appends it to the end of the start command for a number of the different containers (not tomcat). Its perhaps a little naive and simplistic but this gives my users the flexibility they need to specify cli args.

cgfrost:
@johnmcteague We are considering a generic way to override buildpack config through environment variables. This conversation is happening in issue #122 could you look over there to keep it all in one place and as this issue was closed off a while ago. Thanks.

cgfrost:
FYI,

This should now be achievable with the support added in #122 See the readme for more information.
https://github.com/cloudfoundry/java-buildpack#configuration-and-extension

PR #99, opened 2014-09-17
allow hidden remote git url

spaungam:
I have a usage scenario where I'd like to customize the displayed version information by the buildpack and hide the remote git url.

For comparison, here is the sample output of the default behavior where version.yml hasn't been provided
```
-----> Java Buildpack Version: 6d68606 | git@github.com:spaungam/java-buildpack.git#6d68606
```
Here is a sample of a customized `version.yml` with the intention to hide the remote git url:
```
version: '1234'
remote: ''
hash: ''
```

Resulting output of the displayed version information with the customized `version.yml`:
```
-----> Java Buildpack Version: 1234 | #
```

As you can see, there's the unwanted characters `|` and `#`.  This merge request is an adjustment on the version processing to only append these characters if there is a non-empty remote/hash string value.  When the remote/hash string is empty, the desired output is:
```
-----> Java Buildpack Version: 1234
```

### DIFF
--- a/spec/java_buildpack/buildpack_version_spec.rb
+++ b/spec/java_buildpack/buildpack_version_spec.rb
@@ -1,6 +1,6 @@
 # Encoding: utf-8
 # Cloud Foundry Java Buildpack
-# Copyright 2013 the original author or authors.
+# Copyright 2013-2015 the original author or authors.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


jgawor:
Can this change be merged? As @spaungam explained, it's a simple change that does not break anything and does not change the default behavior. And the use case for this change is also pretty straightforward. Since the buildpack can be forked into internal repositories and/or packaged with binaries dependencies, printing the Git url of the buildpack might not always make sense. In fact, sometimes it might be confusing to the user if an internal location of the buildpack is printed.

cgfrost:
We are still not convinced of the value of this pull request against the possibility this will make support harder due to the serviceability information being printed out. Beyond testing and playing around etc...  packaging a buildpack from a git repository that has uncommitted changes is not a good idea. Also, we will never be able to know exactly what will get packaged by a buildpack as this happens when a buildpack is run and not when it's built. That's OK as we're associating a packaged buildpack with a particular git repo and commit only and we value this. I don't understand the use case that requires building a buildpack from inside a git repo with uncommitted changes.

jgawor:
I don't understand why you claim this will make support harder. The default behavior does not change. The only case the Git url would not be displayed is if someone packaged the buildpack with version.yml without the remote and hash properties. In all other (usual) cases it would still be there as before. In the special case, there are other means such as the buildpack version that can be used for support reasons. 
This is not so much about uncommitted changes but about displaying a piece of information that might not be relevant in all cases.

davidcurrie:
Ben/Chris - to take a specific scenario: a PaaS provider clones the buildpack in to a private Git repo so that they can modify the config to pull in a different version of Java. The provider uses the version number as a reference back to their internal commit, and the specific levels of the binaries it pulled in when it was packaged, so that they can provide support to their customers. The remote/hash are meaningless to the consumer of the buildpack and the provider would prefer not to expose details of their internal infrastructure. No uncommitted changes - just non-public information that the provider would rather not share with the end user.

cgfrost:
Thanks @davidcurrie, that helps. We will get back to you early next week.

nebhale:
After @davidcurrie's explanation of the use-case, we're going to accept this contribution.  @spaungam, can you please:

* rebase onto `master`
* ensure that `bundle exec rake` passes without problems
* update the commit message to meet our [Contribution Guidelines](https://github.com/cloudfoundry/java-buildpack/blob/master/CONTRIBUTING.md)

and then resubmit.  Thanks.

spaungam:
@nebhale, the updates to the pull request have been applied per your request.  Thanks for everyone's help to resolve this.

spec/java_buildpack/buildpack_version_spec.rb
```diff
@@ -96,6 +96,28 @@
     expect(buildpack_version.to_hash).to eq({})
   end
 
+  it 'excludes remote string when remote and hash values from config/version.yml are empty',
+     configuration: { 'hash' => '', 'remote' => '', 'version' => 'test-version' } do
+    expect(buildpack_version.to_s).to eq('test-version')
+  end
+
+  it 'excludes include remote string when remote and hash values from config/version.yml are missing',
```
nebhale: Pretty sure that this should be simply `includes remote string when remote and ...`.  Can you please update?

spaungam:
Thanks for catching that.  I've updated the test description per your suggestion.

PR #117, opened 2014-12-04
Fix buildpack_version to work natively in windows

jghiloni:
The packaged output of the java buildpack has the git hash of the most recent baseline appended to the filename. The code that does this makes some unix-centric assumptions that break things when built natively in windows. Specifically:

 * Ruby Pathname objects require the correct path separator, so appending "../../.git" results in a non-existent directory. Using the Pathname.join method resolves the issue
 * To find the git executable, the class calls the native "which" command, and that does not exist in windows. As of Windows 2003, the "where" command exists and is comparable, so use that when on Windows environments.

I've tested this on Windows 8.1 and Ubuntu 14.04 LTS to ensure it didn't break things, with no ill effects noted. Note that the tests didn't complete successfully on Windows, but I think that's more an issue of the tests on Windows rather than the code change itself.

### DIFF
--- a/lib/java_buildpack/buildpack_version.rb
+++ b/lib/java_buildpack/buildpack_version.rb
@@ -94,8 +94,8 @@ module JavaBuildpack
 
     private
 
-    GIT_DIR = (Pathname.new(__FILE__).dirname + '../../.git').freeze
-
+    GIT_DIR = (Pathname.new(__FILE__).dirname.join('..','..','.git')).freeze
+	
     private_constant :GIT_DIR
 
     def git(command)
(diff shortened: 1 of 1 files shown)


cgfrost:
@jghiloni Could you please squash this pull request in to a single commit and also rebase on to master. Once that is done I would like to accept your pull request. I apologise for the slow response.

jghiloni:
Sure, will do.

jghiloni:
Didn't know about rebase and squashing ... Blew away my repo and recreated with the changes in a single commit, will issue a new pull request for that.

jghiloni:
Opened pull request 132

PR #132, opened 2015-01-22
Allow native Windows build support

jghiloni:
Refactor buildpack_version.rb to use native windows path separators
and tools (where instead of which) when in Windows to allow git hashes
to be added to output zip. Same as pull request 117 but rebased.
Issue: 117


cgfrost:
Thanks @jghiloni looks good to me, I have recommended it be accepted. Unless no one else sees a problem this should happen pretty soon.

Issue #122, opened 2015-01-06
Support for user specified MaxPermGen and MaxMetaSpace without fork


youngm:
Over the last couple of years we've gathered a lot of user experience with the java-buildpack and its fork to customize approach.  For the most part this approach has worked well and usually when faced with the choice to fork or just go with the default behaviour our users realize the default behaviour is good enough for them.

However, one issue that seems to frequently pop up is the issue of perm gen tuning.  It is difficult to find a set of permgen settings that works for most smaller apps but don't waste memory for apps that require more permgen.

We could provide our users with small and large permgen forks but we already provide our users with standard jdk7 and jdk8 forks.  Given more permgen flexibility would jump our supported forks from 2 to 4 and potentially add complexity for most of our users that don't need that complexity.

I'd like to propose an option, for apps that need it, to specify a fixed permgen/metaspace size for use in memory heuristics calculations without forking the buildpack.  I'm not proposing a universal solution that would give users fine grained control over all of their memory settings.  From my experience the only item that really seems to need to be tweaked is Permgen/metaspace all our users are happy to let the buildpack calculate the other memory settings.

I was thinking that users today have the option of specifying JAVA_OPTS via an environment variable.  I'd like to propose that if the user supplies an MaxPermGen or MaxMetaSpace option that the buildpack honor that as if it were a fixed `metaspace/permgen` config parameter supplied to the jvm `/config/open_jdk_jre.yml`.  All other permgen and metaspace parameters would be invalid as user specified values as they are today.

Thoughts?

cgfrost:
We will have a conversation about this internally, I can't make a call on it myself.

FYI: https://www.pivotaltracker.com/story/show/86290998

cgfrost:
Hi Mike,

Are you thinking that specifying a minimum permgen/metaspace would be enough or that you want to give an exact value? Also, do you see in the future that you might want to override other memory settings or even other config values. I have some concerns; that we provide a specific solution for metaspace/permgen but then have to generalise it further down the road. I also think it is potentially confusing for users if they do provide an exact and we know it's too high (or low?) to work, do we adjust it and provide some feedback on the console or just let the app go and fail.

No matter what we decide to do I'm in favour of the current behaviour remaining the default so the question is how best to provide additional configuration options.

Chris.

youngm:
[...]

PR #154, opened 2015-03-03
Appdynamic 2.4

hhercules:
What these changes make available for us is the ability to have a top level environment name where we can then group applications, or have a top level application with environment information to separate out apps with same name but different environment. Also this helps us remove AppDynamics license overuse by creating reusable instance names. Also this allows us to view history since name stays consistent.


nebhale:
@hhercules Thank you for the submissions, but we probably won't be accepting either this PR or #155.  Rather than having any fundamental issue with them, they overlap some already scheduled work that we're doing.

First, we're going to be [making `applicationName`, `tierName`, and `nodeName` configurable](https://www.pivotaltracker.com/story/show/89413382) using a precedence system involving the service credentials payload and configuration files.  Secondly, we're going to be making [all configuration available via environment variables](https://github.com/cloudfoundry/java-buildpack/issues/122).

Between these two issues, I believe that we'll add the configurability that you require.  Would you agree?

hhercules:
Yes that will be great, thank you.

Hercules, Christian Hildebrando

cgfrost:
As there have been no more comments for a while, this request is being closed off.

nebhale:
The first bit of this has now gone in: https://github.com/cloudfoundry/java-buildpack/pull/164

nebhale:
The second bit of this has now gone in: https://github.com/cloudfoundry/java-buildpack/issues/122

PR #155, opened 2015-03-03
Appdynamic 2.5

hhercules:
What these changes make available for us is the ability to have a top level environment name where we can then group applications, or have a top level application with environment information to separate out apps with same name but different environment. Also this helps us remove AppDynamics license overuse by creating reusable instance names. Also this allows us to view history since name stays consistent.


cgfrost:
As there have been no more comments for a while, this request is being closed off.

nebhale:
The first bit of this has now gone in: https://github.com/cloudfoundry/java-buildpack/pull/164

nebhale:
The second bit of this has now gone in: https://github.com/cloudfoundry/java-buildpack/issues/122

Issue #157, opened 2015-03-06
JRE memory calculation doesn't take the maximum number of threads in to account


lhotari:
Our application gets killed under load with this error message in events log ```index: 0, reason: CRASHED, exit_description: out of memory, exit_status: 255``` .
After digging in to it, I think it's caused by a bug in the JVM memory calculation. A process in the warden container gets killed by Linux cgroups OOM killer when it goes over it's resource limits. I found the location where the ["out of memory" event is created](https://github.com/cloudfoundry/warden/blob/ad18bff/warden/lib/warden/container/features/mem_limit.rb#L68-L73). I also found the docs for [cgroups memory limits](https://www.kernel.org/doc/Documentation/cgroups/memory.txt). This behaviour is also [defined in the Warden protocol](https://github.com/cloudfoundry/warden/tree/master/warden#limit-handle-mem-value).

Back to the actual issue:

I'd expect that the formula for calculation the total amount of memory is something like this:
```heap + metasize/permgen + native + (thread_stack_size * maximum_number_of_threads)```

However there isn't a documented way to specify the maximum number of threads.
In Tomcat configuration, [the default maxThreads settings is 200](http://tomcat.apache.org/tomcat-8.0-doc/config/executor.html#Standard_Implementation).

There is this comment in [stack_memory_bucket.rb](https://github.com/cloudfoundry/java-buildpack/blob/634b666c497f674db833491cd3fa184d2e661f53/lib/java_buildpack/jre/memory/stack_memory_bucket.rb#L24-L26)
```
   # This class represents a memory bucket for stack memory. This is treated differently to other memory buckets
   # which have absolute sizes since stack memory is specified in terms of the size of an individual stack with no
   # definition of how many stacks may exist.
```

I found a reference to ```num_threads``` in the source code: https://github.com/cloudfoundry/java-buildpack/blob/623c428ed4b8fd411b2368235aa3896e0eef4984/lib/java_buildpack/jre/memory/weight_balancing_memory_heuristic.rb#L110-L116 . This doesn't make sense to me. Is something missing?

I'd like to be able to specify the maximum number of threads in the JRE  memory configuration. The memory calculation could then take this in to account when calculation the memory limits and doing the heuristics.

Maximum thread limits should be adjusted accordingly in the Tomcat configuration so that you don't run in to trouble with the default java-buildpack configuration.

***

Some comments about how Warden handles memory limits:
[...]

PR #158, opened 2015-03-06
Add sensible defaults for Tomcat http connector

lhotari:
- change connectionTimeout to 5000 ms (Tomcat default is 60000ms)
- enable SO_REUSEADDR and TCP_NODELAY by default
- limit maxThreads to 50 (Tomcat default is 200)

### DIFF
--- a/resources/tomcat/conf/server.xml
+++ b/resources/tomcat/conf/server.xml
@@ -19,7 +19,9 @@
 <Server port='-1'>
 
     <Service name='Catalina'>
-        <Connector port='${http.port}' bindOnInit="false"/>
+        <Connector port='${http.port}' bindOnInit="false"
+            connectionTimeout="5000" socket.soReuseAddress="true" socket.tcpNoDelay="true"
+            maxThreads="50" minSpareThreads="10" />
 
         <Engine defaultHost='localhost' name='Catalina'>
             <Valve className="org.apache.catalina.valves.RemoteIpValve" protocolHeader="x-forwarded-proto"/>


nebhale:
@lhotari Have you run these changes past @markt-asf?  I'm not confident in making the change for everyone unless he's on board.

markt-asf:
connectionTimeout="5000"
This also reduces the keep-alive timeout since none is explicitly set. Are you sure that is what you want? Also, 5000 is on the low side for the connection/read time out. It would be fine on a LAN. For mobile clients with a poor connection it could be a little low.

socket.soReuseAddress="true"
What problem are you trying to solve with this setting?

socket.tcpNoDelay="true"
This should be OK but I'd be interested to know if there was any measureable improvement when using this option.

maxThreads="50"
If anything, I'd say 200 was too low a default for production usage. What is the reasoning for lowering to 50?

minSpareThreads="10"
This is already the default so why specify it?

lhotari:
@markt-asf Some context: In the CloudFoundry Warden container, the Java process gets killed without warning when the RSS goes over the limit for the container. The number of active threads is one factor in how much memory the Java process will consume. I started another issue about taking the maximum number in to account when calculating the memory sizes in Java buildpack: #157. That is related to this Tomcat configuration PR. 

`maxThreads="50"`
Each thread will consume more memory. Going to 200 threads will kill the container when there isn't enough memory allocated for 200 threads. Perhaps 50 is a more sensible default for Tomcat in CloudFoundry since we should be allocating memory for the threads too when doing the memory calculation (#157). 
`minSpareThreads="10"`
I just added this to show what the minimum number of threads is.

`connectionTimeout="5000"`
I noticed that the default out-of-box Tomcat config sets `connectionTimeout="20000"` in the server.xml file . I found out that the default is 60000 when you don't specify connectionTImeout. That is certainly too high. Perhaps 5000 is too low, but I found many suggesting that value for webservers. You are right that it might cause problems for mobile clients. 

`socket.soReuseAddress="true"`
We don't need this one.

nebhale:
@lhotari If the number of threads in contributing to an OOM then the solution is not to modify the number of threads, but rather modify the memory allocations to take into account their _actual_ proportion of the memory space.  It seems that if we do that, the only thing left in play on this issue is `connectionTimeout`.

markt-asf:
OK, with no connectionTimeout specified it will default to 60000 (60s) and I agree that is too high. Absent an argument for a different value I'd recommend explicitly setting it to 20000 (20s) to align with Tomcat's out of the box value.

nebhale:
https://www.pivotaltracker.com/story/show/91083074

cgfrost:
Closing without pulling. A smaller change has been made to set the `connectionTimeout` to 20s. After all the discussion this option best suits all of the buildpack users. If there is a special use case that requires different configuration the existing strategy of forking the buildpack should be used.

Commit: 50d9b5f88ccb064aed38641811255c1c8a7e1d0a

PR #164, opened 2015-03-12
Configurable AppDynamics application & node names

cgfrost:
This commit makes it possible to supply values for the AppDynamics
application and node names in the configuration and through service
credentials. Testing and documentation also updated.

[#89413382]


nebhale:
Seems like the documentation is missing changes to to the User Provided Service table.  It needs to contain the keys that can be set for the new features.

PR #168, opened 2015-03-24
User configuration via variables

cgfrost:
This commit adds support for overriding buildpack configuration with
environment variables. The way Java config is processed has been
modified to allow this new support to work when configuring the
required version of Java. Includes new tests and documentation, see
the main README.md for more details.

[#86997400]

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,13 @@ The following are _very_ simple examples for deploying the artifact types that w
 * [Spring Boot CLI](docs/example-spring_boot_cli.md)
 
 ## Configuration and Extension
-The buildpack supports configuration and extension through the use of Git repository forking.  The easiest way to accomplish this is to use [GitHub's forking functionality][] to create a copy of this repository.  Make the required configuration and extension changes in the copy of the repository.  Then specify the URL of the new repository when pushing Cloud Foundry applications.  If the modifications are generally applicable to the Cloud Foundry community, please submit a [pull request][] with the changes.
+The buildpack supports extension through the use of Git repository forking. The easiest way to accomplish this is to use [GitHub's forking functionality][] to create a copy of this repository.  Make the required extension changes in the copy of the repository. Then specify the URL of the new repository when pushing Cloud Foundry applications. If the modifications are generally applicable to the Cloud Foundry community, please submit a [pull request][] with the changes.
+
+Buildpack configuration can be overridden with an environment variable matching the configuration file you wish to override minus the `.yml` extension and with a prefix of `JBP_CONFIG`. The value of the variable should be valid inline yaml. For example, to change the default version of Java to 7 and adjust the memory heuristics apply this environment variable to the application.
+
+```cf set-env my-application JBP_CONFIG_OPEN_JDK_JRE '[version: 1.7.0_+, memory_heuristics: {heap: 85, stack: 10}]'```
+
+Environment variable can also be specified in the applications `manifest` file. See the [Environment Variables][] documentation for more information.
 
 To learn how to configure various properties of the buildpack, follow the "Configuration" links below. More information on extending the buildpack is available [here](docs/extending.md).
 
@@ -125,6 +131,7 @@ This buildpack is released under version 2.0 of the [Apache License][].
 [Cloud Foundry]: http://www.cloudfoundry.com
 [contributor guidelines]: CONTRIBUTING.md
 [disables `remote_downloads`]: docs/extending-caches.md#configuration
+[Environment Variables]: http://docs.cloudfoundry.org/devguide/deploy-apps/manifest.html#env-block
 [GitHub's forking functionality]: https://help.github.com/articles/fork-a-repo
 [Grails]: http://grails.org
 [Groovy]: http://groovy.codehaus.org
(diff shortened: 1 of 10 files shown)


youngm:
Looks good to me though perhaps the `[version: 1.7.0_+]` syntax could use a little more explanation.  How would I specify more than one config entry for a file?

jgawor:
How about using JBP_CONFIG instead of JAVA_BUILDPACK_CONFIG as the environment variable prefix? It's a bit shorter and makes it consistent with the JBP_LOG_LEVEL environment variable.

cgfrost:
@jgawor Good point, I'll make that change. I'll update the example to show a bigger change as well.

README.md
```diff
@@ -24,7 +24,13 @@ The following are _very_ simple examples for deploying the artifact types that w
...
+The buildpack supports extension through the use of Git repository forking. The easiest way to accomplish this is to use [GitHub's forking functionality][] to create a copy of this repository.  Make the required extension changes in the copy of the repository. Then specify the URL of the new repository when pushing Cloud Foundry applications. If the modifications are generally applicable to the Cloud Foundry community, please submit a [pull request][] with the changes.
+
+Buildpack configuration can be overridden with an environment variable matching the configuration file you wish to override minus the `.yml` extension and with a prefix of `JBP_CONFIG`. The value of the variable should be valid inline yaml. For example, to change the default version of Java to 7 and adjust the memory heuristics apply this environment variable to the application.
+
+```cf set-env my-application JBP_CONFIG_OPEN_JDK_JRE '[version: 1.7.0_+, memory_heuristics: {heap: 85, stack: 10}]'```
+
+Environment variable can also be specified in the applications `manifest` file. See [Environment Variables][here] for more information.
```
nebhale: Broken link

lib/java_buildpack/jre/open_jdk_like.rb
```diff
@@ -65,15 +66,26 @@ def release
...
         @droplet.sandbox + 'bin/killjava.sh'
       end
 
       def memory
-        sizes      = @configuration[KEY_MEMORY_SIZES] || {}
-        heuristics = @configuration[KEY_MEMORY_HEURISTICS] || {}
+        sizes      = @configuration[KEY_MEMORY_SIZES] ? @configuration[KEY_MEMORY_SIZES].clone : {}
+        heuristics = @configuration[KEY_MEMORY_HEURISTICS] ? @configuration[KEY_MEMORY_HEURISTICS].clone : {}
+
+        if @version < VERSION_8
```
nebhale: I feel like we've done this before somewhere :)  Did you see another place where this occurs?

cgfrost: We did something similar in the original attempt to detect the Java version by scanning the byte code.

README.md
cgfrost: Which link? They are all working for me...

lib/java_buildpack/util/configuration_utils.rb
```diff
@@ -53,7 +64,33 @@ def load(identifier, should_log = true)
...
+            if hash_v1.key? key
+              hash_v1[key] = do_resolve_value(key, hash_v1[key], value, should_log)
+            else
+              logger.warn { "User config value for '#{key}' is not valid, existing property not present" } if should_log
+            end
+          end
+          hash_v1
+        end
+
+        def do_resolve_value(key, v1, v2, should_log)
```
nebhale: This whole thing is a bit redundant.  Shouldn't you be able to do:

```ruby
v1_hash = v1.is_a? Hash
v2_hash = v2.is_a? Hash

return do_merge v1, v2, should_log if v1_hash && v2_hash
return v2 if !v1_hash && !v2_hash

logger.warn {...}
v1
```

spec/java_buildpack/util/configuration_utils_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+      allow(YAML).to receive(:load_file).and_return('foo' => { 'one' => '1', 'two' => 2 },
+                                                    'bar' => { 'alpha' => { 'one' => 'cat', 'two' => 'dog' } })
+    end
+
+    it 'load configuration file' do
+      expect(described_class.load('test')).to eq('foo' => { 'one' => '1', 'two' => 2 },
+                                                 'bar' => { 'alpha' => { 'one' => 'cat', 'two' => 'dog' } })
+    end
+
+    it 'overlays matching environment variables' do
```
nebhale: This pollutes the environment for all subsequent tests.  You need to use the `:environment` support found [here](https://github.com/cloudfoundry/java-buildpack/blob/master/spec/application_helper.rb#L28-L39) ([example](https://github.com/cloudfoundry/java-buildpack/blob/master/spec/java_buildpack/framework/java_opts_spec.rb#L34)).

lib/java_buildpack/util/configuration_utils.rb
cgfrost: Do you mean the 'do_resolve_value', method. If I remember right, rubocop will complain about the exclamation marks. The previous lines could be tidied up a little.
```return v2 if !v1_hash && !v2_hash```

cgfrost:
Updated

Issue #183, opened 2015-05-05
lib/java_buildpack/repository/repository_index.rb rather tightly coupled /etc/redhat-release 


chrisumbel:
The handling on line `lib/java_buildpack/repository/repository_index.rb:88` for RH-based OSes has some interesting implications. 

1. the tests can't be run on non-centos RH-based OSes directly, which means it can't easily be built on Jenkins, which is fedora-based. This is due to the fact that the string "CentOS" is matched on. I realize we're not targeting a fedora stemcell, but it seems an unnecessary side-effect of tight-coupling. Also, seems silly to require someone to bring up a centos VM or some such to run the tests.
2. tests also can't be run on CentOS 7 because it's format includes the word "Linux" i.e. "CentOS Linux release 7.1.1503". I assume that would need fixed before this buildpack could be used on a CentOS 7 stemcell or am I off base?  
3. easy fix, this would parse "CentOS 10" as "CentOS 1" or "CentOS 20" as "CentOS 2" due to matching on (\d) rather than (\d+). 

I suppose none of these points have any effect whatsoever on production deployments right now, but seemed worth mentioning. If you deem it worth of being looked at I'd be more than willing to get a few commits in to help.

nebhale:
Yeah, I guess we've never tested it on any of these other platforms as they aren't targets for Cloud Foundry itself.  I suspect this will have low priority, but we'll be happy to get to it, and take contributions.

PR #184, opened 2015-05-05
OS Detection and specs work on non-centos and future CentOS versions

chrisumbel:
Non-Centos RH OSes (i.e. Fedora which is used on Jenkins) were similarly unable to test because they got trapped out in repository_index.rb. OS detection specs even on CentOS 6 would fail when doing tests for the non-RH OSes.

OS detection on CentOS 7 would fail completely due to format changes for /etc/redhat-release. 

OS detection would like have detected incorrectly on >9 as only a single digit was matched by regex. 

Integration tests (spec/bin/*) still fail on non-Centos 6, RHEL-based OSes, but all other tests pass now.

Related to #183

### DIFF
--- a/lib/java_buildpack/component/base_component.rb
+++ b/lib/java_buildpack/component/base_component.rb
@@ -121,7 +121,7 @@ module JavaBuildpack
         download(version, uri, name) do |file|
           with_timing "Expanding #{name} to #{target_directory.relative_path_from(@droplet.root)}" do
             FileUtils.mkdir_p target_directory
-            shell "tar xzf #{file.path} -C #{target_directory} --strip 1 2>&1"
+            shell "tar x#{gzipped?(file) ? 'x' : ''}f #{file.path} -C #{target_directory} --strip 1 2>&1"
           end
         end
       end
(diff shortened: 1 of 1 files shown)


nebhale:
@chrisumbel Did you intend to close this issue?  Contribution looks good other than the `rubocop` failures.

chrisumbel:
oh, I was going to address those and re-submit.

nebhale:
Just force push a commit to the same branch and it adds it to this PR directly.  You don't need a separate PR for an update like that.

chrisumbel:
Sounds good. I'll address it shortly.

chrisumbel:
Addressed!

212 files inspected, no offenses detected

nebhale:
@chrisumbel Can you please squash down the two commits into a single one and update the branch one last time?

chrisumbel:
Squashed. Think I did that right.

nebhale:
Sadly, no :)  If you check out the ["Commits" tab](https://github.com/cloudfoundry/java-buildpack/pull/184/commits) you'll see that you've now got 4 commits.  Roughly, you're looking for:

```bash
$ git rebase --interactive origin/master
$ git push chrisumbel master -f
```

(assuming that `origin` is `https://github.com/cloudfoundry/java-buildpack.git`, and `chrisumbel ` is `https://github.com/chrisumbel/java-buildpack.git`)

PR #185, opened 2015-05-05
OS Detection and specs work on non-centos and future CentOS versions

chrisumbel:
Non-Centos RH OSes (i.e. Fedora which is used on Jenkins) were similarly unable to test because they got trapped out in repository_index.rb. OS detection specs even on CentOS 6 would fail when doing tests for the non-RH OSes.

OS detection on CentOS 7 would fail completely due to format changes for /etc/redhat-release.

OS detection would like have detected incorrectly on >9 as only a single digit was matched by regex.

Integration tests (spec/bin/*) still fail on non-Centos 6, RHEL-based OSes, but all other tests pass now.

Related to #183

trying again

### DIFF
--- a/lib/java_buildpack/repository/repository_index.rb
+++ b/lib/java_buildpack/repository/repository_index.rb
@@ -85,7 +85,8 @@ module JavaBuildpack
         redhat_release = Pathname.new('/etc/redhat-release')
 
         if redhat_release.exist?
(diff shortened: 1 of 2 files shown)


chrisumbel:
Redid the pull request as ended up in too funky a state.

chrisumbel:
Thanks, and sorry for the inverse-squash yesterday:)

nebhale:
No worries.  First time is rough for everybody.

PR #193, opened 2015-06-02
Download tar enhancement

cdevarenne:
In method download_tar the logic to detect if the archive is a gzipped file (tar.gz) adds an 'x' to the tar parameters instead of a 'z'. When executed, the command results into "tar xxf" instead of "tar xzf".

The other thing is that the download_tar method currently does not handle archives with bz2 extensions.

This pull request fixes the issue for gzipped archives and adds support for bz2 files (additional handler added bor bz2 files).

### DIFF
--- a/lib/java_buildpack/component/base_component.rb
+++ b/lib/java_buildpack/component/base_component.rb
@@ -178,6 +178,10 @@ module JavaBuildpack
         file.path.end_with? '.gz'
       end
 
+      def bzipped?(file)
+        file.path.end_with? '.bz2'
+      end
+
     end
 
   end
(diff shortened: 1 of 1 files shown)


nebhale:
@cdevarenne Good catch on my mistake in the code.  The one problem I have with this contribution is the duplication of the `shell` line.  I think a better technique is to append a `compression_flag` to the string, in a (better) implementation of what I originally did.  Something like:

```ruby
shell "tar x#{compression_flag(file)}f #{file.path} -C #{target_directory} --strip 1 2>&1"
```

Can you please refactor this submission (or alternately, we can put it on our backlog to do it)?

Thanks again for catching the bug.

cdevarenne:
Thanks for the comment, I will refactor accordingly.

cgfrost:
Hi,

Now we know about this issue we will get it fixed, can you let us know if you intend to resubmit a pull request? Either way we are very grateful for you spotting the issue in the first place.

cdevarenne:
Thank you for the nudge. I re-submitted the pull request, please let me
know if it is all right or needs editing.


Claude

PR #195, opened 2015-06-03
Download tar fix

cdevarenne:
In method download_tar the logic to detect if the archive is a gzipped file (tar.gz) adds an 'x' to the tar parameters instead of a 'z'. When executed, the command results into "tar xxf" instead of "tar xzf".

The other thing is that the download_tar method currently does not handle archives with bz2 extensions.

This pull request fixes the issue for gzipped archives and adds support for bz2 files (additional handler added for bz2 files).


.gitignore
```diff
@@ -2,5 +2,6 @@
 .idea/workspace.xml
 .yardoc
 build/
+.DS_Store
```
cgfrost: This is platform specific and should live in your personal gitignore `$HOME/.gitignore`.

cgfrost:
These changes all look good to me, please can you address the one comment about the ignore file and squash/rebase the pull request in to a single commit. There should be no need to close this pull request and open a new one. Just push the changes to your branch and github will do the rest.

cdevarenne:
Thank you Christopher, I did a rebase/squash and the build passed now.
Thank you for your patience.

Claude

PR #204, opened 2015-07-10
Add default values for heap and stack to support custom fixed values

### LINKED ISSUE #202: Cannot specify stack or heap values for memory_calculator memory_sizes.
Using a value

    JBP_CONFIG_OPEN_JDK_JRE: '[jre: {version: 1.8.0_+}, memory_calculator: {memory_sizes: {metaspace: 64m, heap: 170m, stack: 512k}}]'

Results in the following errors during a push;

      ERR [ConfigurationUtils]             WARN  User config value for 'heap' is not valid, existing property not present
      ERR [ConfigurationUtils]             WARN  User config value for 'stack' is not valid, existing property not present

The documentation lists both heap and stack as parameters for memory_sizes along with metaspace.

leachbj:
To allow fixed values to be set for heap, stack and native as per
https://github.com/cloudfoundry/java-buildpack/blob/master/docs/jre-open_jdk_jre.md#memory-sizes
default values need to be included in the configuration.

[resolves #202]

### DIFF
--- a/config/open_jdk_jre.yml
+++ b/config/open_jdk_jre.yml
@@ -23,8 +23,11 @@ memory_calculator:
   version: 1.+
   repository_root: ! '{default.repository.root}/memory-calculator/{platform}/{architecture}'
   memory_sizes:
+    heap: 64m..
     metaspace: 64m..
+    native: 64m..
     permgen: 64m..
+    stack: 256k..
   memory_heuristics:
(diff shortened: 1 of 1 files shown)


cgfrost:
Hi, after some discussion and investigation we decided to implement this in a different way but it will still achieve the intention of this pull request.

PR #207, opened 2015-07-15
Fixed resources section of dynatrace documentation

jossy:
The resource section seems to be copied from another framework:
- fix naming
- remove profile example since it does not apply to dynatrace


cgfrost:
Thank you for the contribution.

Issue #196, opened 2015-06-12
JCE Unlimited Strength should be put under resources/oracle_jre/jre/lib/security/ in case of using Oracle JRE


devtdeng:
In case os using  Oracle JRE, JCE Unlimited Strength local_policy.jar should be added to   `resources/oracle_jre/jre/lib/security/` instead of `resources/oracle_jre/lib/security`. 

Please correct documentation at
https://github.com/cloudfoundry/java-buildpack/blob/master/docs/jre-oracle_jre.md#additional-resources

nebhale:
Just to double check I've just enumerated the list of files in the Oracle JRE.  I think you'll find that [there is no `jre/` directory](https://gist.github.com/nebhale/49cec4cfa8ee4e1cebee#file-gistfile1-txt-L337-L346) in it.

dmikusa-pivotal:
I think the confusion is that if you download the "server jre", which appears to be structured like a JDK, the directory path is different.

Can we make a note of that in the docs?

```
$ tar tzf server-jre-8u45-linux-x64.gz | grep security
jdk1.8.0_45/jre/lib/security/
jdk1.8.0_45/jre/lib/security/cacerts
jdk1.8.0_45/jre/lib/security/trusted.libraries
jdk1.8.0_45/jre/lib/security/java.policy
jdk1.8.0_45/jre/lib/security/US_export_policy.jar
jdk1.8.0_45/jre/lib/security/blacklist
jdk1.8.0_45/jre/lib/security/blacklisted.certs
jdk1.8.0_45/jre/lib/security/java.security
jdk1.8.0_45/jre/lib/security/local_policy.jar
```

devtdeng:
@nebhale We need to make a note in docs.  If follow current docs at [1], it won't work for Oracle JRE because it require local_policy.jar to be added under resources/oracle_jre/jre/lib/security/ instead of resources/oracle_jre/lib/security.

[1] https://github.com/cloudfoundry/java-buildpack/blob/master/docs/jre-oracle_jre.md#additional-resources

cgfrost:
Could you submit a pull request with the changed docs please. To make sure we get it right.

Thanks.

nebhale:
[...]

PR #230, opened 2015-09-15
document local_policy.jar location for 'server jre'

imod:
as discussed in #196, the documentation about the JCE installation should be enhanced for the 'server jre'


cgfrost:
Thanks for the contribution.

PR #235, opened 2015-09-30
Allow setting of DynaTrace agent name via environment variable

tom-collings:
A customer has asked for a non-default (that is, something different than the application name) to be used for the DynaTrace agent name.  By setting an environment variable, we can allow end users to set the agent name.

The environment variable is JBP_CONFIG_DYNATRACE_AGENT_NAME.  If it exists, the value of this environment variable will be used for the DynaTrace agent name in the JVM Options.  If it does not exist, the default behaviour (using the application name) will be used.

This is my first foray into ruby.  I used other ruby files in the buildpack as a general style guide, any feedback will be greatly appreciated.

Also, its unclear how to sign the CLA.  Should I be receiving some notification with instructions?

### DIFF
--- a/lib/java_buildpack/framework/dyna_trace_agent.rb
+++ b/lib/java_buildpack/framework/dyna_trace_agent.rb
@@ -59,7 +59,7 @@ module JavaBuildpack
       end
 
       def agent_name
-        "#{@application.details['application_name']}_#{profile_name}"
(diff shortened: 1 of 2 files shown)


nebhale:
Completed in https://github.com/cloudfoundry/java-buildpack/commit/6e530398f55a304bff45722a6c90a21e213d2c0c

PR #243, opened 2015-11-01
Enhancements to AppDynamics Config in Java Buildpack

nkatre:
These changes will incorporate the following modifications:

 1. Introduction of variable name **APPDYNAMICS_APPLICATION_NAME** which can be set using `cf set-env` command
 2. Changing the node name which was initially `<instance_index>` of the application to `<tier-name>:<instance_index>` to provide more visibility on tier-node  mapping

To understand all the changes lets consider the following two examples:
 

> **I.**  If **APPDYNAMICS_APPLICATION_NAME** is SET

    cf push abc 

    cf set-env abc APPDYNAMICS_APPLICATION_NAME xyz

|  APPDYNAMICS_APPLICATION_NAME  | Application name | Tier name  | Node name  |
| ------------- | ------------- | ------------ |------------- |
| xyz  | xyz  | abc | abc:0|

> **II.** If **APPDYNAMICS_APPLICATION_NAME** is NOT SET

    cf push abc

|  APPDYNAMICS_APPLICATION_NAME  | Application name | Tier name  | Node name  |
| ------------- | ------------- | ------------ |------------- |
| null  | abc  | abc-tier | abc-tier:0|

### DIFF
--- a/config/app_dynamics_agent.yml
+++ b/config/app_dynamics_agent.yml
@@ -18,5 +18,5 @@
 version: 4.1.+
 repository_root: ! '{default.repository.root}/app-dynamics'
 default_application_name: 
-default_node_name: ! '$(expr "$VCAP_APPLICATION" : ''.*instance_index[": ]*\([[:digit:]]*\).*'')'
-default_tier_name: ! '$(expr "$VCAP_APPLICATION" : ''.*application_name[": ]*\([-_a-zA-Z0-9]*\).*''):$(expr "$VCAP_APPLICATION" : ''.*instance_index[": ]*\([[:digit:]]*\).*'')'
+default_node_name: ! '$(expr "$VCAP_APPLICATION" : ''.*application_name[": ]*\([-_a-zA-Z0-9]*\).*''):$(expr "$VCAP_APPLICATION" : ''.*instance_index[": ]*\([[:digit:]]*\).*'')'
+default_tier_name: ! 
--- a/config/app_dynamics_agent.yml
+++ b/config/app_dynamics_agent.yml
@@ -17,6 +17,6 @@
 ---
 version: 4.1.+
 repository_root: ! '{default.repository.root}/app-dynamics'
-default_application_name: 
+default_application_name:
 default_node_name: ! '$(expr "$VCAP_APPLICATION" : ''.*application_name[": ]*\([-_a-zA-Z0-9]*\).*''):$(expr "$VCAP_APPLICATION" : ''.*instance_index[": ]*\([[:digit:]]*\).*'')'
-default_tier_name: 
+default_tier_name:
--- a/config/app_dynamics_agent.yml
+++ b/config/app_dynamics_agent.yml
@@ -17,6 +17,6 @@
 ---
 version: 4.1.+
 repository_root: ! '{default.repository.root}/app-dynamics'
-default_application_name:
+default_application_name: 
 default_node_name: ! '$(expr "$VCAP_APPLICATION" : ''.*instance_index[": ]*\([[:digit:]]*\).*'')'
-default_tier_name:
+default_tier_name: ! '$(expr "$VCAP_APPLICATION" : ''.*application_name[": ]*\([-_a-zA-Z0-9]*\).*''):$(expr "$VCAP_APPLICATION" : ''.*instance_index[": ]*\([[:digit:]]*\).*'')'
--- a/config/app_dynamics_agent.yml
+++ b/config/app_dynamics_agent.yml
@@ -19,4 +19,4 @@ version: 4.1.+
 repository_root: ! '{default.repository.root}/app-dynamics'
 default_application_name: 
 default_node_name: ! '$(expr "$VCAP_APPLICATION" : ''.*application_name[": ]*\([-_a-zA-Z0-9]*\).*''):$(expr "$VCAP_APPLICATION" : ''.*instance_index[": ]*\([[:digit:]]*\).*'')'
-default_tier_name: ! 
+default_tier_name: 
--- a/spec/java_buildpack/framework/app_dynamics_agent_spec.rb
+++ b/spec/java_buildpack/framework/app_dynamics_agent_spec.rb
@@ -22,7 +22,7 @@ describe JavaBuildpack::Framework::AppDynamicsAgent do
   include_context 'component_helper'
 
   let(:configuration) do
-    { 'default_tier_name'        => nil,
+    { 'default_tier_name'        => "$(expr \"$VCAP_APPLICATION\" : ''.*application_name[\": ]*\\([-_a-zA-Z0-9]*\\).*''):$(expr \"$VCAP_APPLICATION\" : ''.*instance_index[\": ]*\\([[:digit:]]*\\).*'')",
       'default_node_name'        => "$(expr \"$VCAP_APPLICATION\" : '.*instance_index[\": ]*\\([[:digit:]]*\\).*')",
       'default_application_name' => nil }
   end
--- a/spec/java_buildpack/framework/app_dynamics_agent_spec.rb
+++ b/spec/java_buildpack/framework/app_dynamics_agent_spec.rb
@@ -22,7 +22,7 @@ describe JavaBuildpack::Framework::AppDynamicsAgent do
   include_context 'component_helper'
 
   let(:configuration) do
-    { 'default_tier_name'        => "$(expr \"$VCAP_APPLICATION\" : ''.*application_name[\": ]*\\([-_a-zA-Z0-9]*\\).*''):$(expr \"$VCAP_APPLICATION\" : ''.*instance_index[\": ]*\\([[:digit:]]*\\).*'')",
+    { 'default_tier_name'        => nil,
       'default_node_name'        => "$(expr \"$VCAP_APPLICATION\" : '.*instance_index[\": ]*\\([[:digit:]]*\\).*')",
       'default_application_name' => nil }
   end


cgfrost:
Hi @nkatre 

We won't accept the pull request with so many commits, they should be squashed etc... down to a single commit to keep the projects Git history clean.

Introducing an environment variable to provide configuration is against the style of the buildpack. All configuration should go through the config files which can be set with environment variables. I propose a different change, where by the configuration file `default_....` values always win over anything in the credentials payload. This may not be great for when the credentials should win so there is also an option to add a new set of fallback values in the config file that occur when a value isn't given in either the credentials or as a `default_...` value. This will be a much smaller change and stay in keeping with the rest of the buildpack.

pranta:
Hi @cgfrost 

I work with Nikhil, here at AppDynamics.

We agree with your first suggestion of squashing the multiple commits down to a single commit.

With regard to your second suggestion, regarding placing the environment variable APPDYNAMICS_APPLICATION_NAME into a configuration file would not work for us as it would need to be hard-coded. By placing this in the .rb file allows us to dynamically adjust the value depending on what is specified in the credentials. If your argument is to do with the actual name of environment variable, then we are open to renaming it to either DISTRIBUTED_APPLICATION_NAME or MULTI_TIER_APPLICATION_NAME since that more clearly defines the semantics of that variable. In Cloud Foundry, by default each "application" is single-tier. In AppDynamics, by default each "distributed application" is multi-tier. This environment variable allows us to bind multiple Cloud Foundry single-tier applications into a single AppDynamics multi-tier distributed application. If you wish, we could get on a phone call to discuss this further.

nebhale:
@pranta We're unlikely to make any changes to the names of environment variables as everything you want to do is already possible using the existing configuration mechanisms.  At the moment [the application name is already set using the following priority](https://github.com/cloudfoundry/java-buildpack/blob/master/lib/java_buildpack/framework/app_dynamics_agent.rb#L62-L66):

1.  `application-name` specified in the service instance credentials payload (`cf cups`)
2.  `default_application_name` specified either in the `config/app_dynamics_agent.yml` file or by setting the equivalent environment variable (`cf set-env JBP_CONFIG_APP_DYNAMICS_AGENT { default_application_name: <VALUE>}`)
3.  The Cloud Foundry application name (`cf push`)

Your creation of the `APPDYNAMICS_APPLICATION_NAME` environment variable is a direct redundancy of Option 2.

docs/framework-app_dynamics_agent.md
```diff
@@ -17,14 +19,14 @@ When binding AppDynamics using a user-provided service, it must have name or tag
 
 | Name | Description
 | ---- | -----------
```
nebhale: Pretty much none of these changes are correct.  None of `account-access-key`, `account-name`, `application-name`, `port`, `ssl-enabled`, `tier-name` are required.

docs/framework-app_dynamics_agent.md
```diff
@@ -41,6 +43,37 @@ The framework can be configured by modifying the [`config/app_dynamics_agent.yml
 | `repository_root` | The URL of the AppDynamics repository index ([details][repositories]).
 | `version` | The version of AppDynamics to use. Candidate versions can be found in [this listing][].
 
+
```
nebhale: This is unnecessary as it's in the RP and should not be documented in the repository.

lib/java_buildpack/framework/app_dynamics_agent.rb
```diff
@@ -60,9 +60,10 @@ def supports?
       private_constant :FILTER
 
       def application_name(java_opts, credentials)
-        name = credentials['application-name'] || @configuration['default_application_name'] ||
-          @application.details['application_name']
+        name = ENV['APPDYNAMICS_APPLICATION_NAME'] || credentials['application-name']
```
nebhale: This is an unneeded change as the introduction of `APPDYNAMICS_APPLICATION_NAME` is redundant.  Configuration of the the application name can already be done in three other ways.

lib/java_buildpack/framework/app_dynamics_agent.rb
```diff
@@ -83,7 +84,7 @@ def host_name(java_opts, credentials)
 
       def node_name(java_opts, credentials)
         name = credentials['node-name'] || @configuration['default_node_name']
-        java_opts.add_system_property('appdynamics.agent.nodeName', "#{name}")
+        java_opts.add_system_property('appdynamics.agent.nodeName', @appdynamics_tier_name + ":#{name}")
```
nebhale: I'm unclear why the tier name needs to be attached as the node name.  Doesn't the AppDynamics UI already separate nodes by their tiers (in a tree structure)?

docs/framework-app_dynamics_agent.md
nkatre: Hi Ben,

We currently do-not support auto-provisioning mechanism where if the account-access-key, account-name is not specified then we automatically provision an account for such users. Currently such mechanism is not supported by AppDynamics

Hence, parameters such as account-access-key, account-name, port, ssl-enabled, host-name are required.

However, since application-name, tier-name and node-name are automatically populated using the logic mentioned in java-buildpack, these parameters can be set as optional in the doc file.

lib/java_buildpack/framework/app_dynamics_agent.rb
nkatre: I agree. I think `cf set-env <app_name> JBP_CONFIG_APP_DYNAMICS_AGENT '[default_application_name: <custom-name>]'` should be the best way to handle this.

lib/java_buildpack/framework/app_dynamics_agent.rb
nkatre: There are other multiple views within the controller, that only shows node names. It becomes difficult map those nodes to appropriate tiers when we use `instance_index` as node-name.  This is for better readability sake.

docs/framework-app_dynamics_agent.md
nkatre: Agreed !

docs/framework-app_dynamics_agent.md
nebhale: Based on our work against on-prem instances in the past, none of the items listed were required.  There are defaults for `port` and `ssl-enabled`, `account-access-key` and `account-name` are not required for on-prem, and you must only specify a host name.

lib/java_buildpack/framework/app_dynamics_agent.rb
nebhale: @youngm Can you weigh in on this a little.  Do you think this will help visualization in large CF systems?

docs/framework-app_dynamics_agent.md
pranta: @nebhale You're right - prior to AppDynamics Version 4.1 in the minimalist case of a single-tenant on-premise controller, only the controller-host was required and all other parameters were optional or could be derived. However, once Version 4.1 was released, account-access-key has become a mandatory parameter even for single-tenant on-premise controllers.

Spoken to @nkatre about this and he will revert to making all other parameters optional besides controller-host and account-access-key.

pranta:
@nebhale Yes, I wasn't aware of the JBP_CONFIG_APP_DYNAMICS_AGENT environment variable. We can certainly use that for this purpose and revert the addition of the APPDYNAMICS_APPLICATION_NAME environment variable.

lib/java_buildpack/framework/app_dynamics_agent.rb
pranta: @youngm We are waiting for you to weigh in your thoughts.

youngm: @pranta @nebhale Sorry Let me talk with some of my custom system engineers and see what they think.  I'll get back soon.

youngm: @pranta @nebhale So, in our app dynamics service broker we set the node name credential to an expression like:
```
$(expr "$VCAP_APPLICATION" : '.*"application_name[": ]*\\([a-z_0-9-]\\+\\).*')[$(expr "$VCAP_APPLICATION" : '.*"instance_index[": ]*\\([0-9]\\+\\).*')]-[provo]
```

So our node looks like "cf_app_name[0]-[deployment-name]"

I agree with @nkatre that index alone is not a good default for node name.  However, I don't like the idea of hard coding in the BP the prefixing of the Tier name in the Node name.

Why not change the default node name to include the application name similar to the expression I'm using?  Then people can continue to make the node name whatever they want.

Am I following this discussion correctly?

lib/java_buildpack/framework/app_dynamics_agent.rb
```diff
@@ -99,7 +100,14 @@ def ssl_enabled(java_opts, credentials)
       def tier_name(java_opts, credentials)
         name = credentials['tier-name'] || @configuration['default_tier_name'] ||
           @application.details['application_name']
-        java_opts.add_system_property('appdynamics.agent.tierName', "#{name}")
+        @appdynamics_tier_name = ''
+        if @appdynamics_application_name == name
+          java_opts.add_system_property('appdynamics.agent.tierName', "#{name}-tier")
```
youngm: For this change again why not just use an expression as the default that takes the app name that appends -tier?

lib/java_buildpack/framework/app_dynamics_agent.rb
pranta: In our case, the cf_app_name will always have the same value as appdynamics_tier_name.

lib/java_buildpack/framework/app_dynamics_agent.rb
pranta: Same comment here. In our case, the cf_app_name will always have the same value as appdynamics_tier_name.

lib/java_buildpack/framework/app_dynamics_agent.rb
nebhale: Well, that's not strictly true.  At the moment, the tier name can be set using the credentials payload as well.  So the CF application name is _not_ always going to the the same as the AppDynamics tier name.

youngm: Exactly so perhaps you can just change the default to grab the app_name from the runtime environment variable instead of hard coding the conversion in the JBP.

nebhale: Yeah, that's starting to sound like a good alternative.

nkatre: Hi,
We do understand that the node_name is configurable by different ways. The reason why we propose to change the node-name is because  the default behaviour of the buildpack of mapping the node_name to the instance_index would not help customers who completely rely on the default behaviour of the BP OR are not that accustomed to CF CLI OR who are first time users of AppDynamics and after instrumenting their app see the node_name having values as 0,1,2 etc... replicated across all the tiers of their application.

nebhale: @youngm is suggesting a change, which I think is a good one, to add the application name to the default node name.  And he suggests doing this in [`config/app_dynamics_agent.yml`](https://github.com/cloudfoundry/java-buildpack/blob/master/config/app_dynamics_agent.yml#L21) which is the appropriate location.

config/app_dynamics_agent.yml
```diff
@@ -17,6 +17,7 @@
 ---
 version: 4.1.+
 repository_root: ! '{default.repository.root}/app-dynamics'
-default_application_name:
-default_node_name: ! '$(expr "$VCAP_APPLICATION" : ''.*instance_index[": ]*\([[:digit:]]*\).*'')'
+default_application_name: 
+#default_node_name: ! '$(expr "$VCAP_APPLICATION" : ''.*instance_index[": ]*\([[:digit:]]*\).*'')'
```
nebhale: No need for this commented line.

config/app_dynamics_agent.yml
```diff
@@ -17,6 +17,7 @@
...
+default_node_name: ! '$(expr "$VCAP_APPLICATION" : '.*"application_name[": ]*\\([a-z_0-9-]\\+\\).*')[$(expr "$VCAP_APPLICATION" : '.*"instance_index[": ]*\\([0-9]\\+\\).*')]-[provo]'
```
nebhale: Don't need the `-[provo]` on the end here.

docs/framework-app_dynamics_agent.md
nebhale: I believe we're still planning on supporting pre 4.1.  What was the release date for that?

config/app_dynamics_agent.yml
youngm: Someone may want to double check that expr syntax.  My regex skills are passable at best.

nebhale: I see two realistic alternatives:

```
! '$(expr "$VCAP_APPLICATION" : ''.*application_name[": ]*\([-_a-zA-Z0-9]*\).*'')[$(expr "$VCAP_APPLICATION" : ''.*instance_index[": ]*\([[:digit:]]*\).*'')]'
```

which gives you `java-main-application[0]` or 

```
! '$(expr "$VCAP_APPLICATION" : ''.*application_name[": ]*\([-_a-zA-Z0-9]*\).*''):$(expr "$VCAP_APPLICATION" : ''.*instance_index[": ]*\([[:digit:]]*\).*'')'
```

 which gives you `java-main-application:0`.  I prefer the second, but don't have a strong opinion.  Anyone else have an opinion?

config/app_dynamics_agent.yml
```diff
@@ -17,6 +17,6 @@
 ---
 version: 4.1.+
 repository_root: ! '{default.repository.root}/app-dynamics'
-default_application_name:
-default_node_name: ! '$(expr "$VCAP_APPLICATION" : ''.*instance_index[": ]*\([[:digit:]]*\).*'')'
-default_tier_name:
+default_application_name: 
+default_node_name: ! '$(expr "$VCAP_APPLICATION" : ''.*application_name[": ]*\([-_a-zA-Z0-9]*\).*''):$(expr "$VCAP_APPLICATION" : ''.*instance_index[": ]*\([[:digit:]]*\).*'')'
+default_tier_name: !
```
nebhale: I'm pretty sure that this should be an empty line.  What we want is the YAML equivalent of `null` there, not the YAML equivalent of an empty string.

config/app_dynamics_agent.yml
```diff
@@ -17,6 +17,6 @@
 ---
 version: 4.1.+
 repository_root: ! '{default.repository.root}/app-dynamics'
-default_application_name:
-default_node_name: ! '$(expr "$VCAP_APPLICATION" : ''.*instance_index[": ]*\([[:digit:]]*\).*'')'
-default_tier_name:
+default_application_name:
```
nebhale: The empty spaces here and line 22 shouldn't be added.  You're aiming with a diff that only has a single line changed.

nkatre:
Closing this PR and submitting a new pull request with just the required changes https://github.com/cloudfoundry/java-buildpack/pull/245

PR #245, opened 2015-11-09
Changed the value of default node name for AppDynamics

nkatre:
The default node name will now be "app_name:instance_index"


nebhale:
Next time we'd prefer if this stayed attached to the original issue.  As it stands now, the discussion about the changes is in another issue which might as well make it invisible to people investigating in the future.

Issue #161, opened 2015-03-10
killjava.sh not invoked on java.lang.OutOfMemoryError: unable to create new native thread


youngm:
[This stack overflow topic](http://stackoverflow.com/questions/23373117/xxonoutofmemoryerror-doesnt-work-for-java-lang-outofmemoryerror-unable-to-cr) theorizes that the `OnOutOfMemoryError` process cannot be created because there are no more threads available to the OS in this situation.

Perhaps we could handle this case for simple `.war` files by implementing the OOM script in a tomcat valve then invoking `System.exit()` in the case of an OOME so that the process exits and the instance restarts?

Spring Boot apps and such would be on their own in this case but perhaps Spring Boot could provide something general here.

youngm:
Actually I suppose as a valve this may not be very useful.  In the case of creating a thread it is probable that the call trace may be in another thread where the Valve is not in the call chain.  Oh well.  I guess we can log this issue away as interesting data and close it I suppose.

I don't have any ideas of how to mitigate this issue.

youngm:
Perhaps a better solution that may work in more cases than just tomcat Is to create a java agent that sets a default UncaughtExceptionHandler [0] that can find uncaught OOMEs and System.exit().

I recently had another application get an OOME not get handled by OnOutOfMemoryError.  This error message looked like this:
```
OpenJDK 64-Bit Server VM warning: Exception java.lang.OutOfMemoryError occurred dispatching signal SIGTERM to handler- the VM may need to be forcibly terminated
```

Perhaps a default UncaughtExeptionHandler would handle this type of OOME as well?

[0] http://docs.oracle.com/javase/8/docs/api/java/lang/Thread.html#setDefaultUncaughtExceptionHandler-java.lang.Thread.UncaughtExceptionHandler-

cgfrost:
FYI, Issue created https://www.pivotaltracker.com/story/show/90196742

nebhale:
@youngm We've not tried anything in regards to this (and I'm not sure how definitive we could be anyway), but it seems that the OOM situation would preclude the creation of the `Runtime.exec()` infrastructure to execute the script (or equivalent).  Specifically, we'd need to create buffers to capture `stdout` and `stderr` which require more memory.

What's your read on that?

youngm:
[...]

Issue #156, opened 2015-03-05
make OnOutOfMemoryError handling configurable in java-buildpack since killing the process can cause problems for applications that don't have memory leaks


lhotari:
The current solution in java-buildpack kills the running JVM when an OutOfMemoryError happens. This causes problems to some applications that run fine most of the time and don't contain actual memory leaks. The OOMEs are caused by some parts of the application consuming too much memory to produce a report etc. .

I have created [a simple servlet that demonstrates this type of situation](https://github.com/lhotari/hello-jbp-diagnostics/blob/master/src/main/groovy/io/github/lhotari/hello/AllocMemServlet.groovy). You can run it locally with `./gradlew tomcatRun` and access the url http://localhost:8080/hello-jbp-diagnostics/allocmem . 

Without the OnOutOfMemoryError script, the JVM would recover just fine since after the OutOfMemoryError, the thread would exit it's current processing loop and any memory allocated during that processing would be free for GC. That can be demonstrated with the AllocMemServlet sample.

I assume that The OnOutOfMemoryError script would have to do it's work asynchronously and wait a short time period (1-2 seconds) to see if the app recovers after the thread(s) that received the OutOfMemoryError have exited their processing loops and released the allocated memory for GC.

lhotari:
It would be possible to add custom OutOfMemoryError handling that happens inside the JVM with [Thread.setDefaultUncaughtExceptionHandler](http://docs.oracle.com/javase/7/docs/api/java/lang/Thread.html#setDefaultUncaughtExceptionHandler(java.lang.Thread.UncaughtExceptionHandler)) . That logic could take care of tracking whether the OOM is "real" by keeping track of the rate of OOMEs. I'm not sure if it's possible to use the OnOutOfMemoryError JVM option together with the UncaughtExceptionHandler. Just some implementation ideas for the improved "memory leak" detection that seems to be the original reason to add the killjava.sh solution.

cgfrost:
Do you have a genuine use case where an application throws an OOM legitimately. This really shouldn't happen and I'm inclined to say the problem is in the application or it just needs to be assigned more memory.

lhotari:
@cgfrost In most cases a Java application and the JVM is able to recover from an OOME that is caused by a thread allocating too much memory in the current execution stack. The JVM usually recovers from this type of situation when the OOME happens since the memory allocated from objects in the execution stack will get released after the OOME.

[...]

PR #246, opened 2015-11-10
Add support for JVMKill agent in Java Buildpack

rafaelri:
Provide the ability to use JVMKill agent 
(https://github.com/airlift/jvmkill) instead of the regular shell script
that was invoked as an OnOutOfMemoryError handler. This change aims to
make possible for the JVM to recover when the application spawns too
much threads (circumstance that prevents the OnOutOfMemoryError handler
from running).

Issue: #161


nebhale:
I've been looking over #156 and #161 and clearly they're related, but not entirely overlapping.  If we're thinking about changing the OOM behavior, why don't we use this PR as a place to discuss what the final behavior should be, combining both of them.

I think our best option is to replace the existing shell script completely since we know that using `-XX:OnOutOfMemoryError` has memory/threading issues.  The question then is what we want to happen when an OOM is triggered.  It seems like in some circumstances we might not want to exit the JVM, in others we would, and would probably need some configuration around when to do that.  In the case do we want to exit the JVM, do we want to preserve the output from the current script (we can't really do the after, but we could dump the before if it was useful).  Finally, do we want to dump the heap even though by default, most applications don't have enough room in their containers to dump the heap and even if they did, it'd be recycled immediately.

We've got fair amount of Go knowledge on the team (see the [`memory-calculator`](https://github.com/cloudfoundry/java-buildpack-memory-calculator) for an example), so I'd guess that we'd use what we see in [airlift's stuff](https://github.com/airlift/jvmkill) but implement it more robustly in Go. 

/cc @cgfrost @youngm @lhotari @gberche-orange @Zteve

lhotari:
@nebhale +1, for most of the plan. 
Is there other projects using Go for JVMTI agents? It might require a lot of pioneer work if there aren't examples of implementing a JVMTI agent with Go.

nebhale:
Agents are implemented by [`include`-ing a specific header file](https://github.com/airlift/jvmkill/blob/master/jvmkill.c#L21).  Once compiled they're the same statically-linked machine code that they would have been had they been written in raw C.  We'll get to test that, but I don't foresee any problems along those lines.

lhotari:
@nebhale sounds interesting :+1:

nebhale:
@lhotari Any thoughts on what heuristics to use when deciding whether to kill the JVM or proceed?

kelapure:
@nebhale  FWIW I believe killing the JVM and NOT allowing it to proceed is the way to go here. I have never seen any good come about by having these zombie JVMs hang around.

nebhale:
@kelapure Please see https://github.com/cloudfoundry/java-buildpack/issues/156 for examples where killing isn't necessary.

rafaelri:
@nebhale and also reimplementing in Go avoids having (ideally I guess) airlift author to sign the CLA as well. In regard to the heuristics... let me add my 2 cents... what about having something in the lines of what I did here: https://github.com/rafaelri/jvmkill/blob/counter/jvmkill.c : having a threshold of X resource exhausted events in the last Z seconds ?

nebhale:
Yeah, I was just trying to bait someone into putting that into writing as part of this issue.  It's a good stake in the ground, so let's see what others suggest. Also `jvmkill` is [Apache Licensed](https://github.com/airlift/jvmkill/blob/master/LICENSE), so I'm not too worried about using their code.

rafaelri:
@nebhale in fact it isn't even that hard to build it in go:

    package main

    // #cgo CFLAGS: -I"/usr/lib/jvm/jdk-8-oracle-x64/include" -I"/usr/lib/jvm/jdk-8-oracle-x64/include/linux"
    // #include <jvmti.h>
    import "C"
    import (
	"fmt"
	"unsafe"
    )

    func main() {}

    //export Agent_OnLoad
    func Agent_OnLoad(vm *C.JavaVM, options *C.char, reserved unsafe.Pointer) C.jint {
	fmt.Printf("Hello, world.\n")
	return C.JNI_OK
    }

and build it with `go build -buildmode=c-shared -o libjvmkill.so cloudfoundry.org/gojvmkill`

I'll try to port the code tomorrow but I've already tested and the JVM loaded the Go agent just fine.

lhotari:
@nebhale @rafaelri :+1:  to the simple idea of "having a threshold of X resource exhausted events in the last Z seconds" .

lhotari:
Here's a link to the JVMTI ResourceExhausted event documentation: http://docs.oracle.com/javase/7/docs/platform/jvmti/jvmti.html#ResourceExhausted

rafaelri:
@nebhale only one thing that is worth mentioning although I guess it isn't that relevant: the agent went from 9kb to 3,2mb (something expected due to the Go Runtime embedded in the so). We can consider this increase irrelevant to the footprint of the cf instance or is is something we should be concerned?

nebhale:
@rafaelri At runtime we don't very much mind it's size, but we'll tar it up for downloads which should shrink it back to about ~500kb like the [`memory-calculator`](https://download.run.pivotal.io/memory-calculator/trusty/x86_64/memory-calculator-2.0.1_RELEASE.tar.gz).

rafaelri:
@nebhale regarding the limited capacity for heap dumps... why not provide a histogram of the heap using jmap? For sure it isn't as detailed as a heap dump but it already provides some hint on a possible leak, its output is like the one below:
    rafaelri@samsung550:~/gojvmkill$ jmap -histo:live 9235

     num     #instances         #bytes  class name
    ----------------------------------------------
       1:           985          85504  [C
       2:           450          51456  java.lang.Class
       3:             8          24984  [B
       4:           501          24880  [Ljava.lang.Object;
       5:           964          23136  java.lang.String
       6:            74           5328  java.lang.reflect.Field
       7:           256           4096  java.lang.Integer
       8:            93           3720  java.lang.ref.SoftReference
       9:           111           3552  java.util.Hashtable$Entry
      10:             8           3008  java.lang.Thread
      11:            86           2952  [I
      12:            59           2184  [Ljava.lang.String;
      13:            40           1920  sun.util.locale.LocaleObjectCache$CacheEntry
      14:            55           1760  java.util.concurrent.ConcurrentHashMap$Node
      15:            21           1344  java.net.URL

rafaelri:
@nebhale I am progressing with the go agent but things started to get messy after I've stumbled upon two cgo limitations:
The first prevented me from calling directly the JVMTI functions, leading to the necessity of having code in the preamble:

> Calling C function pointers is currently not supported, 

Source: https://golang.org/cmd/cgo/#hdr-Go_references_to_C

> Note: you can't define any C functions in preamble if you're using exports.

Source: https://github.com/golang/go/wiki/cgo

And I am trying to figure out a proper approach of registering my Go function as a handler on the JVMTI callback. My next attempt is to pass an unsafe.Pointer that points to my go function. My first attempt wasn't promising but I probably did something wrong. I'll keep you updated on this.

nebhale:
I'm going to have our Go expert, @zteve look at this today.

nebhale:
@rafaelri The histogram is an interesting idea, but it seems like it runs into the same `fork()` problem if we try and run it, even via the agent.

lhotari:
@nebhale @rafaelri There is a JVMTI agent that prints out a histogram on github: https://github.com/Cue/polarbear . The licensing of that code seems to be complicated.

nebhale:
Hmm...  Wonder if that's CDDL.  Let me see what I can find without looking at the code.  At least no one else should look at the code for now just to keep a white room separation.

nebhale:
Sun code is MIT licensed, Cue code is ASL licensed.  We're good.  Everyone can take a look now.

lhotari:
Seems to be based on JVMTI example: http://hg.openjdk.java.net/jdk8u/jdk8u-dev/jdk/file/tip/src/share/demo/jvmti/heapViewer/heapViewer.c

Zteve:
@rafaelri Hi. Ben has asked me to join this.

We have decided that Go is probably not going to give us any real advantages over C. It appears that to get the Go executable loaded as an agent we need to use quite a bit of C-code anyway, and there are also Go runtime threads which we'll get whether we want them or not. Since we're not using any of Go's cool features, and since we'd have to do some serious research to get the Go functions registered as callbacks (even indirectly) we think we should stick to C for the time being.

A fork of the JVMKill agent would be a good starting point. I look forward to see what you can come up with, if you want to submit a commit or two.

rafaelri:
@Zteve sure, give me a few days and I'll finish the changes I already started on the agent! I'll only be unavailable between 13th and 17th of November since I'll be travelling. @nebhale in regard to the fork issue in some quick tests here I noticed that it is problematic only in the situation of having too much live threads, with an exhausted heap I don't see any difficulties in spawning jmap and dumping the histogram but the ulimit thing related to the threads is a challenge, I even tried something as having a "prelaunched shell script" only waiting for a keystroke or something to move on with the clean up thing but bash itself forks when running executables from a shell script so I'd have to have more than a "single spare thread" (which in turn can't be guaranteed). 
About the heuristics for killing and its configuration, @nebhale have you been able to see how I have provided the configuration for the "restart component" in my commit? Is it the recommended approach? Do you have any general suggestions/tips/comments?

nebhale:
@rafaelri We're going to maintain a fork of `jvmkill` at https://github.com/cloudfoundry/jvmkill.  When you've got something to take a look at, you should be able to submit a PR there.  I don't have a strong enough background in JVMTI and C in general, so I'm going to ask @Zteve to be your guide as to what we expect for submission.  He'll be able to answer questions such as the recommended way to configuration the agent and integrate the heap view from [the sample code](http://hg.openjdk.java.net/jdk8u/jdk8u-dev/jdk/file/tip/src/share/demo/jvmti/heapViewer/heapViewer.c).

I will be happy to help you get the agent that you've produced into the buildpack.  I think we've got some time until that's ready to go though, so lets focus on the agent first with @Zteve.

rafaelri:
@nebhale makes sense to start with the agent parameters, I'll get back to how to pass them to the agent later on. @Zteve I must admit that I am no expert in C or Go either but this agent thing is definitely within what I can do. I have already a working version using getsubopt for passing the parameters since I wasn't willing to do any string manipulation in C. Do you see any drawbacks in using getsubopt for parsing the parameters and having them comma separated?

rafaelri:
@nebhale should we keep our discussion over here or move back to the issue #161 ?

rafaelri:
@nebhale I've already managed to have a thread dump when the Threads are exhausted by checking the flags passed to the handler function (if it is either JVMTI_RESOURCE_EXHAUSTED_OOM_ERROR || JVMTI_RESOURCE_EXHAUSTED_JAVA_HEAP or JVMTI_RESOURCE_EXHAUSTED_OOM_ERROR || JVMTI_RESOURCE_EXHAUSTED_THREADS)
and issuing a kill with SIGQUIT (that triggers a Thread dump). The next thing I'll attempt is to issue the Heap Dump, I'll try to port the agent @lhotari mentioned.

rafaelri:
@Zteve in regard on how to test our agent. I was thinking about having a Go application load the library and testing it but I think we would still have to move the "kill action" from the resourceExhausted method in order to make it easier for testing.

nebhale:
@rafaelri Let's keep the discussion here, but remember that the code changes will need to exist in a RP in our fork of `jvmkill`.

rafaelri:
@nebhale I was porting the code from the agent @lhotari suggested when I noticed that the effort of implementing the autokill on Polarbear agent would be much less than porting the thread dump and histogram to C on top of jvmkill. Do you see any problems in changing our strategy? Any licensing issue for example? I did a proof of concept here and it seemed easier (I was having a hard time porting parts of the code, the thread dump is almost working except that there's something weird in the part that fetches the line numbers for the dump and I'd still have to port the histogram that seems more complex). If you agree with this strategy, please create a cloudfoundry fork as you did with jvmkill and I'll fork it.

rafaelri:
Actually it isn't a matter of porting the kill mechanism only but I'd also need to strip down the Polarbear agent to its bare minimum (removing a TCP shell) but still it seems lot easier.

rafaelri:
It is already working. The next step is to port the threshold to the Polarbear agent. It is available at https://github.com/rafaelri/polarbear/tree/autokill-threshold

Zteve:
@rafaelri I wonder if you could separate the tailorable kill mechanism from the dump summary feature, so as to make two (smaller) pull requests. The dependencies for the features are different.

I've had a go at running your implementation of the kill-threshold, and have cherry-picked your commit into a branch `kill-threashold-trials`, which you might look at.  Most of the changes I made were to the makefile so as to run the tests more smoothly. Take a look.

I would appreciate a better description of the _purpose_ of the `tests.c` code so as to be able to validate it.

The Java code `JvmKillTestThreads.java`, and my own attempt `jvmKillTest2.java` are not clearly enough testing the right resource exhaustion, nor verifying the threshold rules.  Could we have a more carefully constructed Java test program that can be tailored to run out of memory and/or threads in certain ways to provoke (or not provoke) the killer?

Well done for tackling this, btw: I think we are approaching a solution to a difficult problem.

rafaelri:
Hi @Zteve, I saw your changes to the Makefile, in fact it makes it a lot easier to run the individual test groups. In regards to the tests.c: its purpose was exactly to test the scenarios you imagined in the improved java tests. I thought it was easier and more predictable to test those aspects by invoking directly the agent library and having scenarios in which I would reach the threshold and others that wouldn't reach. I think we can have something similar with a Java application but I thought testing like this would be easier. Perhaps I could provide more comments in the tests.c file and let clear its purpose? And what's your opinion about having this kill+threshold mechanism ported to the Polarbear (as I did on https://github.com/rafaelri/polarbear/tree/autokill-threshold) instead of porting the thread dump+heap histogram code from the Polarbear agent to the Airlift jvmkill?

Zteve:
@rafaelri Thanks. Yes I see the purpose of `tests.c`, and you are right: it is easier to invoke the entry points from another C routine.  These are a good basis for “unit tests” as you are essentially mocking the JVM.  The Java programs would be in the nature of “integration tests”, since they run in a real JVM (or several different ones, maybe?). We should, of course, have both.

I haven't looked at the histogram code very closely because I am keen to get the threshold feature implemented first.  If we are to adopt a version of jvmkill I'm not too bothered about its provenance (properly attributed of course), but I would like to see the features delivered separately. We would certainly want to have different tests for the two aspects. Are you happy to consolidate the kill+threshold part first?

rafaelri:
@Zteve I completely agree with you in regard to having the kill first but there is still one thing that I am not comfortable. Sorry to bother you with such a bureaucratic topic but license wise if we move on with this repository (airlift) and later on we bring changes from Polarbear we will end up with code that is BSD and Apache licensed. On the other hand if we fork from Polarbear and implement the kill mechanism we will have it BSD only since a single "kill" line can't be treated as a derivative work (and the threshold mechanism was ours). If you feel like it is fine to move on with this repo I am fine also. Perhaps we should ask @nebhale opinions as he brought the topic before? We could also have the Polarbear modified agent have the dump options defaulted to off in order to avoid any risks (of not properly testing these features).

rafaelri:
@Zteve I just noticed that we definitely need some locking mechanism in the resourceExhausted callback. Polarbear used the Java CreateRawMonitor but I think it is a complete overkill since we won't share this lock with any Java code and another drawback is that it'll make testing a pain. I've used http://en.cppreference.com/w/cpp/thread/lock_guard and it is working fine (no more random crashes upon Heap Histogram)

rafaelri:
@Zteve btw, the command line parameters and the defaulting to off is already implemented in my polarbear branch.

Zteve:
@rafaelri Sorry for the delay. The licence situation is OK either way, but if you like a Polarbear mod more, then go for it. I think technically it poses no problems for us.

Please modify this PR to have a valid commit that provides *just* the threshold kill mechanism in (cut-down?) Polarbear (if that is way you want to go). Then I can review  it, and subject it to the tests we have lined up (we have a Java program all set to go that runs a tailorable configuration of thread and/or memory exhaustions with time delays).

Don't worry about the makefile changes, just provide the minimal complete threshold kill function. Defer the dump summary output for now.  The `tests.c` program or equivalent is a secondary consideration at this stage; but please include what you've got: we can improve that later.

This is promising work. Here are a few suggestions to speed up acceptance:

1. Ensure you comment all external interfaces properly.
1. Output some help when the code is executed 'standalone' (if it can be).
1. Somewhere document the semantics of all the parameters, including the units used (ms/µs/s?).

I'm looking forward to including this.

Zteve:
@raphaelri Please ensure that the lock is applied only when it is necessary
(I'm not sure we need it for the kill/no kill decision).

Steve Powell          [M: +*44*-*7815*-*838*-*558* H:+*44*-*1962*-*775*-
*598*]
  *Every great art must be supplemented by leisurely discussion, by
stargazing, if you will, about the nature of things.*
Plato *Phædrus*
pivotal.io <http://www.pivotal.io>|SpringSource
<http://www.springsource.org>|RabbitMQ <http://www.rabbitmq.com>

rafaelri:
@Zteve let's do it over the airlift code/repo then, it really makes no difference for me. Later on we'll only change the build to invoke g++ instead of gcc cause I think it doesn't make much sense to port all the classes on Polarbear code to C structures and functions. We won't have much gains. 
And I'll address all your suggestions (comments, tests, ...).
BR

rafaelri:
@Zteve I feel like it needs on the counter code cause I could notice that the thread that runs the resourceExhausted is exactly the same (Java thread) that was allocating resources. And bear in mind that this lock will only "cause retention" after a resourceExhausted event is invoked. It won't pose any performance impact (IMHO).

rafaelri:
@Zteve previous comment wasn't complete... hence it runs on the thread we can have a scenario where multiple Java threads are allocating resources and they all trigger the resourceExhausted event at the same time thus incrementing concurrently the counter variable.

Zteve:
Hmm: I'm losing track of where the comments go. Better here, I think.

Glyn Normington made the following comment on line 49 of `jvmkill.c`:

| _If the desired behaviour is that no more than count_threshold OOMs may occur in any period of 
| time_threshold seconds, then this code doesn't appear to achieve that. For instance, with a 
| time_threshold of 10 and count_threshold of 1, if 2 OOMs occur when `time(NULL)' is 99 and 100, 
| respectively, then it will appear as if the OOMs occurred in distinct 10-second intervals and the JVM 
| will not be killed when it should have been._

rafaelri:
@Zteve in fact there's this corner case. But I sincerely don't think that in our usage scenario this will make much difference. We are more worried about not killing a JVM when it might recover than being strict on this counter, again, IMHO. In the case the application was misbehaving another OOM would probably happen in the time lapse of the second. I opted for this implementation to avoid anything more sophisticated than this (and avoiding having to store the event along a timestamp or having a timer or a thread to periodically reset the timer). With this implementation we have constant memory usage and execution time (sure we give up on some accuracy for the "x events in z seconds" but I feel like it is reasonable).

rafaelri:
@Zteve please have a look at https://github.com/rafaelri/jvmkill/tree/store-events-threshold . Glyn's comments made me think whether we could improve a little bit on the implementation and be more accurate in regard to the event timestamps. I also improved on the comments as you requested but I didn't feel like it makes much sense to have an option to print usage on the agent since we won't have a clear place to point the user to invoke such an option (I even testes here having a help parameter to the agent but I think it would end up only duplicating what we could have on the README for example).

Zteve:
@rafaelri Thanks for those updates. I do think it is important to be “accurate” in the way we treat event counting in an interval.  However, your solution is unnecessarily complex. It is quite possible to have a pre-allocated `count` sized array of time-stamps, and O(1) algorithms for adding a new event and checking for overflow in the time interval. It is not necessary to have a `deque`, which might involve memory allocations/deallocations at an awkward time, nor is it ever necessary to prune the array of entries (and a good job, too, since `pop_back()` while scanning the `deque` backwards by index, feels potentially dangerous).  I'll leave you to discover this, and make the necessary changes.

The lock will only be necessary if the JVMTI protocol doesn't serialize calls to `resourceExhausted`. Can you verify this?  (I'm supposing that C++ releases the lock in the destructor.)

House style will probably not tolerate too many line comments (if the variable/field is named correctly then renaming it in the line comment should be unnecessary), but this can be tidied up along with minor layout concerns. I'm going to get the Java testing running against your version while you work on these very minor concerns.

rafaelri:
@Zteve I had random crashes without the mutex on polarbear code with both thread dumps and heap histograms (which makes me think the JVM does not serialize resourceExhausted callback otherwise the mutex would make no difference) but with the code I am pushing now it made no difference (I've coded a new test hoping to have a concurrent allocation, see the new JvmKillTestThreadsParallel class). And yes, c++ releases the mutex in the destructor. 
In regard to the deque I've changed it to a simpler implementation using arrays as you suggested, please have a look. 
I hope the comments are better (and only where necessary) now.
BR

rafaelri:
@Zteve changes on the same branch as the previous deque impl.

rafaelri:
@Zteve I am also still thinking about changing getTimeMillis to use http://linux.die.net/man/3/clock_gettime with CLOCK_MONOTONIC since I sincerely don't know what might happen if system clock is changed (I know it is ALMOST impossible to happen inside a CF instance but god know what kind of code someone might upload there, isn't it?).

rafaelri:
@Zteve only to let you know that I've changed the getTimeMillis function to use the clock_gettime with CLOCK_MONOTONIC (seems much safer). I've also already updated the branch for the pull request on the jvmkill repo. Code with the arrays is indeed much simpler but one last note, the reverse iterator on the deque wouldn't be affected by operations on its ends, operations in the middle are the ones that invalidates all iterators and the reason I started searching backwards whas that I knew that I was pushing new events in the front and the old ones would accumulate in the end.
Waiting for your feedback to move on with this change.
BR

rafaelri:
Hi @Zteve , any news on this?

nebhale:
Sorry @rafaelri.  This appears to have gotten dropped during the holiday season.  @Zteve will take a look tomorrow.

rafaelri:
@nebhale Great! I am awaiting for any feedbacks regarding the last changes.

rafaelri:
@nebhale Couldn't we start the Ruby part to speed things up?

rafaelri:
hi @nebhale, no news on this?

youngm:
Just wanted to add my voice to this.  I'd love for this PR to get some priority.  It will do great things for Java on CF for my org.

glyn:
Created tracker [feature 114206679](https://www.pivotaltracker.com/story/show/114206679) to cover this PR.

glyn:
Hi @rafaelri. I started looking at this PR.

There's a _lot_ of context that I haven't yet absorbed, so I thought I'd start by running the tests and I got failures because some downloads (e.g. https://raw.githubusercontent.com/rafaelri/jvmkill/master/binaries/x86_64/index.yml) are not available. Should these be available?

(Please note I am offically working on the Java buildpack on Mondays only, so apologies for any delay in responding to you.)

rafaelri:
@glyn I have recreated my fork after Ben created a fork under Cloud Foundry repo to hold our changes and since then we were focusing on the agent first but it is only a matter of rebuilding the agent and zipping. I can upload them today.

rafaelri:
And also I guess that in the end it won't be hosted on github also. Where is the usual place for storing this files related to Java buildpack?

glyn:
@rafaelri Java buildpack dependencies are stored in S3. See [java-buildpack-dependency-builder](https://github.com/cloudfoundry/java-buildpack-dependency-builder) for more information (including a pointer to the S3 indices for `jvmkill`).

rafaelri:
@glyn if you need any help (apart anything related to Mac OS as I said before) please let me know.

glyn:
Thanks @rafaelri. I'm off doing other things until next Monday. Meanwhile @nebhale is going to have a think about how to break down the work into a sequence of small pieces.

glyn:
@rafaelri Hi. I've just started on the story to print a memory histogram on kill (https://www.pivotaltracker.com/story/show/114796249) and am wondering where your latest code is. I've found some commits in https://github.com/rafaelri/polarbear. I haven't yet dug in to polarbear. If you can provide more information to help me make progress, that would be great.

glyn:
@rafaelri In particular, is the intention that when the `printHeapHisto` option is specified, that a histogram is printed on each OOM or only when we decide to kill the JVM?

rafaelri:
hi @glyn  we haven't settled this. I was about to implement only before it got killed but you've raised an interesting point. From my experience it'd be better to have it on each OOM since it'd point the overall status of the memory all the way from the first OOM up to the point the JVM turned not viable and had to be killed. Perharps we should see what @nebhale and @lhotari thinks.

glyn:
Thanks @rafaelri. The current design of `jvmkill` has split two classes out from the main logic with corresponding interfaces. The `Heuristic` interface decides _when_ to act on OOM and the `Action` interface decides _what_ action to take. So we could add the histogram printing to the `Heuristic` if we want it every time or we could add it to the `Action` if we want it just before we kill the JVM. Meanwhile, I'll keep exploring the histogram implementation because that will be the same either way.

rafaelri:
@glyn the code is definitely much cleaner now, the refactoring was incredible. I was thinking whether the printing (heapHisto and threads) should go into the heuristics since I (IMHO) see both as actions. Then we would have to change the Heuristics to have a list of observers (actions) and perharps two events, a regular OOM and something like a critical OOM making it possible to have two Print actions (PrintHistoOnEveryOOM and PrintHistoBeforeKill). Am I making it too complex?

glyn:
@rafaelri Perhaps! What difference do you think there might be between the two histograms (printed by PrintHistoOnEveryOOM and PrintHistoBeforeKill)?

glyn:
@rafaelri Note that the current split of responsibilities between the `Heuristic` and `Action` interfaces is not cast in concrete so we can change it as necessary or completely redesign the whole thing if that's best. _If_ we want to print the same type of histogram on every OOM, we could, for instance, allow there to be multiple heuristics registered. The existing one (`Threshold`) returns `true` when a kill is required and `false` otherwise. A new heuristic (`Histogram`) could print a histogram and always return `false`.

rafaelri:
@glyn (1st message) no difference from the output at all. Only a matter of letting the user choose between a more verbose print on every OOM and the less verbose print before killing only. From my experience I'd always prefer having all of them since usually I look for the trend on the heap to try to detect which situation in the code let the JVM to the get out of heap (and sincerely usually the tread dump helps a lot also but that's another topic).

rafaelri:
@glyn for now I wouldn't change the design. It is rather very clear, expressive and really fits our problem. After having another look at the code I think we could have jvmkill.c++ have a list of actions instead of a single one and notify them (as you suggested) passing the boolean kill flag as parameter. We would only have to take care of having the kill action itself as the last registered action. Then, if we had only printHisto configured we'd only have the printHisto and kill action but if we had all of the action flags on we would end up having three actions registerd (histo, threads and kill). What do you think?

glyn:
@rafaelri Something like that should be ok. The main complication is the parsing of options which is currently delegated to the `Threshold` constructor. We are moving to a design in which more than one helper class is interested in options, so we'll want some way for the helper classes to take a look at the options and say which ones they have consumed. Perhaps the cleanest way would be to pass each option to each of the helper classes via a new `Options` interface with a single method returning a `bool` saying whether the option was consumed or not. Then options which were not consumed by any helper class would be diagnoses as invalid.

rafaelri:
@glyn I agree with you in having a central place for parsing the options I couldn't only understand what would be the possible implications of having more than a single Action consuming an option.

glyn:
@rafaelri something like a future `debug` option, for instance, might apply to all Heuristics/Actions.

I am running out of time today and won't be back to working on this until 25th April since I am on vacation next Monday. Would you have any time in the next two weeks to create a pull request against jvmkill for the histogram function? If so, I could pick it up on the 25th and carry on from where you left off. If you don't have time to do the full thing, feel free to do a partial implementation (e.g. keeping the option parsing as today and hard-coding histograms as always on) which I could then extend.

Regards, Glyn

rafaelri:
@glyn For sure I can give it a try!

rafaelri:
@glyn I've coded the Parameters as a struct (and I am thinking about passing it by copy to avoid any unwanted manipulation but we can change it to a class, protect the elements and pass by reference if you prefer). I have already integrated the polarbear agent code as is but I am planning to do some refactorings part of its code.

glyn:
Thanks @rafaelri. If you can make whatever code you have written available, preferably as a PR, before Monday, I'll be able to take a look at what you've been up to.

rafaelri:
@glyn sure. I'll do the PR today. We still need to improve a few things but
it is already working and with a decent test coverage.
Em 21/04/2016 06:05, "Glyn Normington" <notifications@github.com> escreveu:

glyn:
Wow, I'll look forward to trying this out. Thanks @rafaelfi!

On Thu, Apr 21, 2016 at 2:45 PM, Rafael de Albuquerque Ribeiro <
notifications@github.com> wrote:

> @glyn sure. I'll do the PR today. We still need to improve a few things but
> it is already working and with a decent test coverage.
> Em 21/04/2016 06:05, "Glyn Normington" <notifications@github.com>
> escreveu:
>
> > Thanks @rafaelri <https://github.com/rafaelri>. If you can make whatever
> > code you have written available, preferably as a PR, before Monday, I'll
> be
> > able to take a look at what you've been up to.
> >
> > —
> > You are receiving this because you were mentioned.
> > Reply to this email directly or view it on GitHub
> > <
> https://github.com/cloudfoundry/java-buildpack/pull/246#issuecomment-212821876
> >
>
> >
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/cloudfoundry/java-buildpack/pull/246#issuecomment-212925731>
>



-- 
Regards,
Glyn

kelapure:
I can help test this feature with a malicious app. The app can simulate various kinds of memory leaks.

Sent from my iPhone,
Rohit

glyn:
@kelapure Thanks Rohit! Please take a look at the tests that are currently there (JvmKillTest*.java in the root of this repository), which exhaust heap and threads. What other kinds of memory leaks do you have in mind?

rafaelri:
@glyn and @kelapure I've used this https://github.com/rafaelri/webapp-oom on CloudFoundry while I had access to its free version.

rafaelri:
@glyn do I have to squash my commits into a single one before doing the PR or do you prefer multiple commits?
Another two things I forgot to ask you:
1) I've removed the dependency on some of the tests to the main build as I felt like the tests could run independent of the main build and this made it faster for testing. Do you see any problems that I might have not noticed?
2) I have not coded any tests for the polarbear code itself and I feel like it needs a good refactoring (although it is already working) before we consider this done and unfortunately I don't think like I'll have much time before the next month to help you on this (even though I really wished I had time).

glyn:
@rafaelri: I don't mind multiple commits at all. Sometimes multiple commits
are preferable because it shows the stages of evolution of the code. As for
(1), this sounds ok in principle and I'll look more closely on Monday. As
for (2), I'm happy to spend time refactoring if you can get something which
(a) works and (b) I can at least understand. So if there's anything
particularly subtle or tricky, a comment somewhere would be appreciated,
e.g. in the PR.

On Thu, Apr 21, 2016 at 5:45 PM, Rafael de Albuquerque Ribeiro <
notifications@github.com> wrote:

> @glyn <https://github.com/glyn> do I have to squash my commits into a
> single one before doing the PR or do you prefer multiple commits?
> Another two things I forgot to ask you:
> 1) I've removed the dependency on some of the tests to the main build as I
> felt like the tests could run independent of the main build and this made
> it faster for testing. Do you see any problems that I might have not
> noticed?
> 2) I have not coded any tests for the polarbear code itself and I feel
> like it needs a good refactoring (although it is already working) before we
> consider this done and unfortunately I don't think like I'll have much time
> before the next month to help you on this (even though I really wished I
> had time).
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/cloudfoundry/java-buildpack/pull/246#issuecomment-213006982>
>



-- 
Regards,
Glyn

rafaelri:
@glyn I really asked about the multiple commits thinking about this. And regarding polarbear code it is good and readable but I guess we can improve in regard to a GlobalData struct that was already present in the code and that only now I noticed that I forgot to remove things that I've added myself in the prior attempt of integrating this code into the original jvmkill agent. I'll try to update the PR here.

kelapure:
Native memory leaks. Where JNI library leaks memory.

Sent from my iPhone,
Rohit

rafaelri:
@glyn there is one last aspect I forgot to discuss with you: we need further capabilities on the jvm before doing the heap histogram and for now I've placed this code in the HeapHistogramAction and it is even in a suboptimal place (should have been on the constructor). But the point is that I am unsure whether we should only claim those capabilities if we are going to use it or whether we should ask for all of them ahead of time in the jvmkill.c++ itself and never bother again with this. To be 100% sure we should have to dig a lot on JVM code/documentation in order to know if those capabilities adds any kind of overhead to the JVM but I am pretty sure that we will have a hard time searching for this information.

glyn:
@rafaelri Let's keep the capability obtaining code where it is for now and
do something different if there is a measurable impact on performance.

On Thu, Apr 21, 2016 at 6:37 PM, Rafael de Albuquerque Ribeiro <
notifications@github.com> wrote:

> @glyn <https://github.com/glyn> there is one last aspect I forgot to
> discuss with you: we need further capabilities on the jvm before doing the
> heap histogram and for now I've placed this code in the HeapHistogramAction
> and it is even in a suboptimal place (should have been on the constructor).
> But the point is that I am unsure whether we should only claim those
> capabilities if we are going to use it or whether we should ask for all of
> them ahead of time in the jvmkill.c++ itself and never bother again with
> this. To be 100% sure we should have to dig a lot on JVM code/documentation
> in order to know if those capabilities adds any kind of overhead to the JVM
> but I am pretty sure that we will have a hard time searching for this
> information.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/cloudfoundry/java-buildpack/pull/246#issuecomment-213031355>
>



-- 
Regards,
Glyn

glyn:
@rafaelri - I changed my mind and moved the capability code to the constructor on the basis that if we can't get the necessary capabilities, construction should fail rather than storing up this failure for an awkward surprise during OOM processing. I added some code today to test this, although the tests are not complete and @cgfrost will continue them later in the week. Then the plan is to traverse the live objects on the heap using [FollowReferences](http://docs.oracle.com/javase/7/docs/platform/jvmti/jvmti.html#FollowReferences) with a suitable tag (we need to investigate the precise rules for choosing a tag to be sure of correct behaviour). An entry will be made in a hashmap for the class of each object encountered. The map key will be the class name. The map value will be a structure containing the number of objects of this class and the size of each object. After traversing the heap, we'll then need to sort the map entries and print out a histogram.

glyn:
@rafaelri /cc @cgfrost : I pushed a squashed commit of the heap histogram action constructor tests. I've kept the capabilities to the minimum set -- can_tag_objects -- as per the JVMTI documentation. Not sure about coding standards for static variables: perhaps the static variables named "Mock*" in the test should be named "mock*" instead. That's it from me for now. Over to @cgfrost to pick off the first part of the heap traversal code...

glyn:
@rafaelri /cc @cgfrost : my code on the `live-heap-traversal` branch (off of the main `PolarBear-ReWrite` branch) is now ready for the live heap traversal code (in `heaphistogramaction.c++`) and unit tests (in `heaphistogramactiontests.c++`) to be added. There are the beginnings of mock `HeapStats` in the unit test code, so it should be possible to write this code in isolation from the JVM. OTOH it may be necessary to experiment with the JVM to fill in any gaps in the JVMTI documentation. See the comment two above for the plan for how to traverse live objects...

cgfrost:
@rafaelri /cc @glyn I'm working on a branch called `heap-stats-impl`. Filling out the tests first.

glyn:
Question about the use of tags: http://stackoverflow.com/questions/37439576/are-object-tags-set-by-the-jvm-visible-to-jvmti-agents

glyn:
@hhoffstaette pointed out this for Java 9: https://bugs.openjdk.java.net/browse/JDK-8152669

glyn:
@robharrop mentioned http://www.docjar.com/html/api/sun/tools/attach/HotSpotVirtualMachine.java.html

See heapHisto...

glyn:
@robharrop also mentioned a high dynamic range histogram as an efficient, although not absolutely accurate storage mechanism. This might turn out to be useful if we are ever faced with building the heap histogram with a limited amount of C heap.

youngm:
Where are we at with this PR?  I was really hoping to get some of this functionality sometime and thought the PR was making great progress.

rafaelri:
Probably @nebhale is the best one to answer but as far as I can tell we
have already implemented the kill mechanism and also the heap histogram
code in the agent. I think @nebhale has already deployed the binaries and
now we need to move on with the ruby code for the buildpack but as I said
let's wait for the official answer.
BR

Em 27 de jul de 2016 14:58, "Mike Youngstrom" <notifications@github.com>
escreveu:

nebhale:
The `jvmkill` code is ready to go, it's just finding the time to integrate it into the buildpack.  I'm trying to fit in all of the Concourse work to automate delivery to the repository for buildpack consumption, but it's getting squeezed out by SpringOne prep.  If I don't get it done before the conference, it'll happen shortly after.

There is some question about when it'll go out in a tagged release though.  I'm giving serious thought to putting it, and major changes to the memory calculator into `master`, but doing a long test period before a 4.0 release.

@youngm what do you use as the base for your buildpacks `master` or tagged releases?

youngm:
@nebhale I use tagged releases but when ready I could do some testing with master.

That update sounds great.  I can be patient.  I just wanted to make sure this PR wasn't dead.  Thanks!

glyn:
@youngm Just wanted to check you can see the story in tracker for integrating `jvmkill` into the buildpack: https://www.pivotaltracker.com/story/show/114796351. You can see where it stands in the backlog and this should give you some assurance that it will eventually be worked on! Hope that helps.

youngm:
@glyn yes, I can see it.  I'll keep track of the progress there.  Thanks!

PR #250, opened 2015-11-13
Update open_jdk_jre.yml

spraveenj:


nebhale:
I don't think you meant to submit a PR for this.

cgfrost:
This should be left as an open range so the latest 8.x version of Java is used every time. If you want to lock it down to a specific version please look at the configuration docs on how to set this value using an environment variable. https://github.com/cloudfoundry/java-buildpack#configuration-and-extension

PR #251, opened 2015-11-13
Upgrading gemfire version to 8.2

sbawaska:
Upgraded to latest GemFire release to be used for session caching.

### DIFF
--- a/config/tomcat.yml
+++ b/config/tomcat.yml
@@ -37,16 +37,16 @@ redis_store:
   connection_pool_size: 2
 gemfire_store:
   gemfire:
-    version: 8.0.+
+    version: 8.2.+
     repository_root: ! '{default.repository.root}/gem-fire'
   gemfire_modules:
-    version: 8.0.+
+    version: 8.2.+
     repository_root: ! '{default.repository.root}/gem-fire-modules'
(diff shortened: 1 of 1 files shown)


cgfrost:
Hi,

Thanks for the contribution, can you sign the CLA and I will merge it in. It looks like you work for Pivotal though so if you are in any of the Pivotal Orgs you should just be able to make your membership public and be approved automatically.

sbawaska:
I have made my "pivotal-cf" membership public.

nebhale:
@cgfrost Should we actually slacken this restriction to `8.+`?  It seems like you've uploaded a couple of subsequent releases that we not actually picked up by users.

sbawaska:
The problem with that is if the build-pack picks up a GemFire release, before the service team. Since GemFire servers are backwards compatible, an old client in the buildpack will be able to talk to a new server, but if the buildpack picks up a new client, it won't be able to talk to an old server.

nebhale:
But that's true now as we upgrade to 8.2 as well.  There's no way to guarantee that any user will have upgraded when the buildpack does.  Until the client is backwards compatible, there's no real way to solve this problem.  In lieu of a solution, we go for consistency which is latest possible version at all times.

Issue #257, opened 2015-11-20
no_proxy environment varbiable is ignored when downloading cache for resources behind a firewall for offline buildpacks


samwlui:
Currently all resources to be cached for offline buildpack creation is retrieved using the proxy configured in the http_proxy or https_proxy environment variable. In order to build a buildpack that contains a resource behind a firewall, the no_proxy environment variable needs to be considered so that the proxy can be bypassed.

nebhale:
Closed in lieu of #258.

PR #258, opened 2015-11-21
Added check to consider no_proxy env variable for offline caching

samwlui:
In order to build a buildpack that contains a resource behind a firewall, the no_proxy environment variable needs to be considered so that the proxy can be bypassed. Currently all resources to be cached for offline buildpack creation is retrieved using the proxy configured in the http_proxy or https_proxy environment variable. Added a check to validate whether the uri is listed in the no_proxy env variable which will then pass nill as the proxy host to bypass the proxy.

I have also emailed the signed CLA

Issue: #257

### DIFF
--- a/lib/java_buildpack/util/cache/download_cache.rb
+++ b/lib/java_buildpack/util/cache/download_cache.rb
@@ -233,6 +233,7 @@ module JavaBuildpack
         end
 
         def proxy(uri)
+          env_no_proxy = ENV['no_proxy'] || ENV['NO_PROXY']
           proxy_uri = if secure?(uri)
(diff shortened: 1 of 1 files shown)


nebhale:
@samwlui You'll need to include tests like the ones for `http_proxy` as well as an update to [the documentation](https://github.com/cloudfoundry/java-buildpack/blob/f20749099750b441ebb18e79e3b8a07a7e61f45b/docs/extending-repositories.md#proxies)

nebhale:
Since there's been no response on this issue in a couple of weeks, I'm going to close it.  If you'd like to see it re-opened, please comment on the issue and I'll reopen it.

PR #274, opened 2016-02-19
Extension point for supplying an external Tomcat configuration

violetagg:
Hi,

I would like to propose an extension point for supplying a Tomcat configuration without forking the buildpack.

The new component is disabled by default and can be enabled in case an external Tomcat configuration will be used.

The mechanism for overlaying the configuration will not be changed.
1. Default Tomcat configuration
2. Tomcat configuration provided by the buildpack
3. External Tomcat configuration if it is enabled.

The external Tomcat configuration must be supplied in TAR format and must follow the Tomcat archive structure.

The external Tomcat configuration will not be modified so the supplier should guarantee that the symlinks are allowed.

I would be glad to hear your opinion.

Thanks,
Violeta


testobsessed:
&nbsp;<b>Code Climate</b> has <a href="https://codeclimate.com/repos/5224adaec7f3a3415107004c/pull/274">analyzed this pull request</a>.

cgfrost:
I'm happy with the code changes.

nebhale:
Resolved with #b2bdeb877b620decaca68284eb27fa30ec3507e2

violetagg:
Thanks a lot
Violeta

PR #276, opened 2016-03-01
Improve suspend documentation to include -u none

youngm:
You can only debug suspend if you disable container health checking.  This also has the happy side effect of never timing out the startup of the application in this case.


testobsessed:
&nbsp;<b>Code Climate</b> has <a href="https://codeclimate.com/repos/5224adaec7f3a3415107004c/pull/276">analyzed this pull request</a>.

PR #278, opened 2016-03-07
Add Ruxit agent framework for monitoring 

aloismayr:
This PR adds support for Ruxit monitoring to CF Java apps and includes docs, tests and config. It requires a ´ruxit´ service being bound to an app in order to add the Ruxit agent to JAVA_OPTS agentpath parameter. The default repository points to download.ruxit.com and is maintained by Dynatrace/Ruxit.
The CLA will be in Ben Hale's inbox today.


testobsessed:
&nbsp;<b>Code Climate</b> has <a href="https://codeclimate.com/repos/5224adaec7f3a3415107004c/pull/278">analyzed this pull request</a>.

nebhale:
@aloismayr Can you give me some insight into the contents of the agent ZIP file?  I'm looking for an explanation of what is in the `bin/` directory, but I'd like to get a better idea all the way around.

nebhale:
I'd also like a pointer to some documentation on how the agent is configured if you can provide that as well.

aloismayr:
@nebhale You can take a look into this ZIP file http://download.ruxit.com/agent/paas/unix/java/ruxit-Agent-Linux-1.87.124.zip 
The `lib` and `lib64` directories contain the loader for the agent and some native code that we use to instrument the process. In `bin` there is the part of the agent that is specifically tied to Java.

nebhale:
Yeah, I've take a look at the internals of the ZIP file, which is why I wanted a deeper understanding of what exactly each piece was used for.  Specifically, it appears that the `-javaagent` only refers to one of the `.so`s, so I wanted to understand what the 1.2M ZIP file in `/bin` was used for.

aloismayr:
The part in `/bin` covers things like monitoring JVM memory metrics, JMX metrics, etc. and is being loaded by either the `lib` or the `lib64` so.

nebhale:
How does the `lib` determine the location of those things?  Is it relative to a working directory or relative to the library itself?  Also, you've exposed some configuration via environment variables.  Is there any documentation for the values that can be configuring using that strategy?

aloismayr:
The libraries in `lib` determine the location of those things by relative paths to the library itself, plus the information in `conf/binaries.json` to get right managed and native parts. Our CI makes sure a single ZIP bundle is self-contained and valid so that the loader can find those things. 

In CF we make use of three environment variables to specify the names of some entities which otherwise the agent would determine and set.

`RUXIT_APPLICATIONID` is name of the webapplication (in case it is a web application, otherwise this will be ignored). This usually is auto-detected but we set this to `application_name`. Refer https://help.ruxit.com/pages/viewpage.action?pageId=18612610

`RUXIT_CLUSTER_ID` is the name of the process group. Ruxit groups similar processes that belong to the same application/services and type, e.g., 4 tomcats running the same services would be grouped to one process group. Again, this usually is auto-detected but for CF we set this to the `application_name` as this matches our model of process groups. In case you do a `cf scale APP_NAME -i 4` all the 4 app-instances would belong to the same process group. Refer https://help.ruxit.com/pages/viewpage.action?pageId=8061067 

`RUXIT_HOST_ID` overrides the detected hostname in Ruxit. In CF this normally is the name of the Garden container the app instance is running but the container name isn't meaningful to humans.

nebhale:
@aloismayr Still looking this over.  Just wanted to let you know that I haven't forgotten.

nebhale:
@aloismayr Is there documentation for the `agentpath` configuration arguments?

nebhale:
@aloismayr Also, [this documentation](https://help.ruxit.com/pages/viewpage.action?pageId=8061067) seems to indicate that we might want `RUXIT_NODE_ID` instead of `RUXIT_HOST_ID`.  What are thoughts on this?

aloismayr:
@nebhale No worries! There is no other documentation for `agentpath` parameter than in the buildpack's [`docs`](https://github.com/cloudfoundry/java-buildpack/blob/ruxit/docs/framework-ruxit_agent.md). Normally, Ruxit users don't need to deal with that.

`server` is the endpoint the agent reports data.
`tenant` is the customer's environment id.
`tenanttoken` is the secret needed to report data to the environment.

nebhale:
Sounds good.  Since I can't [find any mention](https://www.google.com/search?client=safari&rls=en&q=RUXIT_HOST_ID&ie=UTF-8&oe=UTF-8) of `RUXIT_HOST_ID`, I just wanted to confirm.

PR #286, opened 2016-04-01
Minor improvement to Ruxit Agent deployment

aloismayr:
With this improvement the buildpack no longer depends on a specific directory structure in the agent ZIP file. All contents in `agent` subdirectory will be moved to droplet.


nebhale:
@aloismayr I just did a little poking about, and it appears this is the result of the fact that the agent zip is now tailored for this deployment?  There's nothing that _shouldn't_ be there any more?

aloismayr:
@nebhale Exactly! The zip only contains what's really needed for the buildpack.

PR #289, opened 2016-04-06
Enable Dynatrace 6.3 Agent and public repo

akirasoft:
Added public repository containing Dynatrace Agent 6.3 and incremented agent version to 6.3

Previously Dynatrace support was baked into the buildpack but commented out due to lack of Dynatrace provided agent download location. We have addressed this. 

We have a valid/signed CLA through the Dynatrace organization.


akirasoft:
Publicized my org membership!

PR #295, opened 2016-05-03
Adding space to the app name shown in NewRelic

bbyers-pivotal:
When the NewRelic service is bound to an app name that exists in other spaces as well for the same account, it is confusing looking at NewRelic because it would lump all those instances into one app on NewRelic. With this change, application names will include the space deployed as well.

Old way
In Cloud Foundry Space/Appname -> Shown in NewRelic
MySpace1/myapp -> myapp
MySpace2/myapp -> myapp

New way
In Cloud Foundry Space/Appname -> Shown in NewRelic
MySpace1/myapp -> MySpace1-myapp
MySpace2/myapp -> MySpace2-myapp


nebhale:
@bbyers-pivotal Have you run this past the New Relic team?  They were involved heavily in the current design (choosing the names for example), and thought this would be the ideal situation.  We've added the ability for these values to be overridden with environment variables for individual applications if they didn't work, but we'd want to run any changes to this past New Relic before merging.

bbyers-pivotal:
@nebhale I have not talked with the New Relic team yet. Who should I contact on the New Relic team to run this past them? Thanks!

nebhale:
After private discussion, this issue will be changed to a documentation update indicating how to achieve a per-app configuration for this functionality.  Alternately, it can be closed completely.

PR #317, opened 2016-08-09
Update dynatrace detection criteria

aloismayr:
Update criteria for detection of dynatrace/ruxit extensions.

### DIFF
--- a/lib/java_buildpack/framework/dyna_trace_agent.rb
+++ b/lib/java_buildpack/framework/dyna_trace_agent.rb
@@ -41,7 +41,9 @@ module JavaBuildpack
 
       # (see JavaBuildpack::Component::VersionedDependencyComponent#supports?)
       def supports?
-        @application.services.one_service? FILTER, 'server'
+        (@application.services.one_service? FILTER, 'server') &&
+        !(@application.services.one_service? FILTER, 'tenant') &&
+        !(@application.services.one_service? FILTER, 'tenanttoken')
       end
 
       private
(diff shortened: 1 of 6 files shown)


nebhale:
@aloismayr This doesn't lead to some overlap between the two frameworks?  They're still mutually exclusive of one another?

Also, can you please update the documentation with the new detection criteria so that there's a human-readable version of the same changes?

lib/java_buildpack/framework/dyna_trace_agent.rb
```diff
@@ -41,7 +41,7 @@ def release
 
       # (see JavaBuildpack::Component::VersionedDependencyComponent#supports?)
       def supports?
-        @application.services.one_service? FILTER, 'server'
+        (@application.services.one_service? FILTER, 'server') && !(@application.services.one_service? FILTER, 'tenant')
```
nebhale: This filter should mirror the positive filter from the Ruxit framework; it needs to require both `tenant` _and_ `tenanttoken`.

aloismayr:
@nebhale You're right. We needed to change the detection criterion for both products (Appmon & Ruxit) because Dynatrace Ruxit recently became Dynatrace. See here: https://blog.ruxit.com/dynatrace-ruxit-becomes-dynatrace/ 
I'll update the documentation accordingly by end of today. Sorry, forgot to do so yesterday.

aloismayr:
@nebhale Any news on this? FYI - Updated the documentation regarding the detection criteria for both products.

nebhale:
Sorry, thought I'd merged that in.  Guess I never pushed.  Give me 10 minutes.

PR #339, opened 2016-10-13
AppDynamics Docs Update

nkatre:
The `account-access-key`, `account-name` are credentials to your controller. And it is a required parameter.
Its highly recommended that one specifies the remaining parameters except the `node-name`

### DIFF
--- a/docs/framework-app_dynamics_agent.md
+++ b/docs/framework-app_dynamics_agent.md
@@ -15,18 +15,18 @@ Tags are printed to standard output by the buildpack detect script
 ## User-Provided Service
 When binding AppDynamics using a user-provided service, it must have name or tag with `app-dynamics` or `appdynamics` in it. The credential payload can contain the following entries:
 
-`NOTE:` ***** marked fields are `Optional` for java agent versions prior to 4.2. However, for agent versions 4.2 and above, these fields are `Required`. Please check `/config/app_dynamics_agent.yml` file for appropriate version.
+`NOTE:` For agent versions 4.2 and above, * marked fields are no longer `Optional`, they are `Required`. Check `/config/app_dynamics_agent.yml` file for appropriate agent version.
 
 | Name | Description
 | ---- | -----------
-| `account-access-key` *| The account access key to use when authenticating with the controller
-| `account-name` *| The account name to use when authenticating with the controller
-| `application-name` *| the application's name
+| `account-access-key`* | (Optional) The account access key to use when authenticating with the controller
+| `account-name`* | (Optional) The account name to use when authenticating with the controller
+| `application-name`* | (Optional) the application's name
 | `host-name` | The controller host name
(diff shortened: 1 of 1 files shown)


nebhale:
Actually, `account-access-key` and `account-name` are not required.  If you are connecting to an on-premise installation, these are not required arguments.  In addition there is a default [`application-name`](https://github.com/cloudfoundry/java-buildpack/blob/master/lib/java_buildpack/framework/app_dynamics_agent.rb#L64), a default [`port`](https://github.com/cloudfoundry/java-buildpack/blob/master/lib/java_buildpack/framework/app_dynamics_agent.rb#L91) (optional parameter to agent), a default [`ssl-enabled`](https://github.com/cloudfoundry/java-buildpack/blob/master/lib/java_buildpack/framework/app_dynamics_agent.rb#L96) (optional parameter to agent), and a default [`tier-name`](https://github.com/cloudfoundry/java-buildpack/blob/master/lib/java_buildpack/framework/app_dynamics_agent.rb#L101).

peterholditch:
The port is not optional.  The agent will not stat successfully unless it is specified.

nebhale:
If that's true, then the change needs to be isolated to that specific update.  In addition to a documentation update, you'll also need to update the code to enforce that fact.

nkatre:
In [this](https://docs.appdynamics.com/display/PRO42/Java+Agent+Configuration+Properties) 4.2.x version documentation, the `account-name`, `account-access-key`, `controller port`, `app name` and `tier name` all are required parameters except node-name.

We previously had received bug/doc update tickets regarding the same, and we are trying to educate most customers with it. But most of them use the default java bp for documentation, hence the need

nebhale:
The buildpack currently supports Agents back to `3.3.2`.  I'm believe the reason why these properties are still optional is that they are not required throughout the entire supported history.

nkatre:
We can add a `NOTE` in the documentation mentioning to refer to AppDynamics docs for the required parameters depending on the version of the agent that they are downloading.

Moreover, the default behaviour of the AppDynamics config file within this bp suggests that the bp downloads the latest 4.2+ agent, which suggests we should somewhat update the literature with a NOTE if not complete modifications. This would be useful for customers.

nebhale:
The default is to use 4.2, but at any time, any application can `cf set-env <APP-NAME> JBP_CONFIG_APP_DYNAMICS_AGENT '{version: 3.2.3}'` and use that version.  In fact, given the installed versions on-prem of AppDynamics that I've seen in customers recently, I'd suggest that changing that configuration to be compatible with their footprint is pretty common.

I'm happy to review a contribution that describes optionality more specifically, without removing it completely.

nkatre:
@nebhale I have proposed a few changes to handle the recent updates

nebhale:
The removal of the `(Optional)` label isn't going to work.  It's convention with the rest of the documentation and will need to stay.

PR #348, opened 2016-11-23
update open_jdk_jre version changed

stgitmobility:
update open_jdk_jre version changed to version 31


nebhale:
I do tho believe that you intended to submit this as a RP.

PR #362, opened 2016-12-15
manifest.yml

rajnishpandey:
added manifest yaml.


nebhale:
I do not believe that you intended to submit this as a contribution.

PR #363, opened 2016-12-16
Update the JRebel framework.

poolik:
Use the 7.x release which has logging enabled by default, so no need to explictly set it. Also 'lib/jrebel.jar' doesn't exist anymore so removed it from the expectations.


cgfrost:
LGTM and packages locally with JRebel 7.0.1 just fine.

cgfrost:
Thank you for the contribution Tõnis.

PR #371, opened 2017-01-24
Improvements for HA for the Luna Security Provider.

jpatchell:
* Adding support for HA logging.
  * Adding HASynchronize section for HA groups.
  * Fixing reconnect attempts option and changing to max retries.

### DIFF
--- a/docs/framework-luna_security_provider.md
+++ b/docs/framework-luna_security_provider.md
@@ -92,6 +92,7 @@ The framework can be configured by modifying the [`config/luna_security_provider
 | Name | Description
 | ---- | -----------
 | `logging_enabled` | Whether to enable the logging wrapper for the Luna Security Provider.  Defaults to `false`.
+| `ha_logging_enabled` | Whether to enable HA logging for the Luna Security Provider.  Defaults to `true`.
 | `repository_root` | The URL of the Luna Security Provider repository index ([details][repositories]).
 | `version` | Version of the Luna Security Provider to use.
 
--- a/lib/java_buildpack/framework/luna_security_provider.rb
+++ b/lib/java_buildpack/framework/luna_security_provider.rb
@@ -110,6 +110,10 @@ module JavaBuildpack
         @configuration['logging_enabled']
       end
 
+      def ha_logging?
+        @configuration['ha_logging_enabled']
+      end
+
       def padded_index(index)
         index.to_s.rjust(2, '0')
       end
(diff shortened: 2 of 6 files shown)


docs/framework-luna_security_provider.md
```diff
@@ -92,6 +92,7 @@ The framework can be configured by modifying the [`config/luna_security_provider
 | Name | Description
 | ---- | -----------
 | `logging_enabled` | Whether to enable the logging wrapper for the Luna Security Provider.  Defaults to `false`.
+| `ha_logging_enabled` | Whether to enable HA logging for the Luna Security Provider.  Defaults to `true`.
```
nebhale: In general, we'd alphabetize these (for my OCD), but I'll do it in a following commit.

lib/java_buildpack/framework/luna_security_provider.rb
```diff
@@ -144,20 +148,27 @@ def write_configuration(servers, groups)
...
-      def write_epilogue(f)
+      def write_epilogue(f, groups)
         f.write <<EOS
 }
 
 HAConfiguration = {
   AutoReconnectInterval = 60;
   HAOnly = 1;
-  ReconnAtt = 20;
+  reconnAtt = -1;
```
nebhale: Does this mean try to reconnect forever?  Is this correct?

jpatchell: Yup, this means that it will attempt to reconnect forever (practically) to recover the member which is a good strategy.

docs/framework-luna_security_provider.md
jpatchell: sounds good!

lib/java_buildpack/framework/luna_security_provider.rb
nebhale: Where this is problematic is if the configuration points to something that will never exist.  It means that it will retry forever (also, not sure what the backoff strategy is) without ever failing.  We generally consider it a bad practice to retry without limit.  But if you all say it's good, I'll let it stand.

jpatchell: The more important change is ReconnectAtt to reconnectAtt as the default value is 0 which means it will never attempt to reconnect (setting is case sensitive).  I will bounce your concerns about having it retry forever off a few people to get some more opinions.

obstacleman: Probably it would be nice for it to be configurable but most people use the -1. (forever) setting.  What guidance do you give as best practices?

nebhale: I don't have a strong opinion on how long, so long as it isn't unbounded.  Each service is unique and their retry characteristics are as well.  I mean, if you haven't been able to connect for 5 minutes, an hour, 6 hours, etc., how do HSM users typically decide that they're never going to be able to connect?  You might also take a look into what HSM behaviors you're trying to protect against; is it just intermittent network hiccups (then retry for a minute or two) or extended downtime due to system updates (then retry an hour or two).

PR #376, opened 2017-02-07
Update dyadic_ekm_security_provider.rb

peer85:
Make ha_mode_standby be one when using Pivotal (turn off auto randomize)


peer85:
10x

PR #414, opened 2017-04-19
Add a framework for the thin launcher

dsyer:
* detects the ThinJarWrapper main class
* adds a lib directory if missing (so java_main container works)
* resolves dependencies on "compile" so the cost is not incurred on startup

TODO: modify spring_boot container to detect a Spring Boot app if it is
there somehow


dsyer:
Probably more work needed here, but it would be useful to know if it's on the right track.

nebhale:
I need to dig into how exactly the thin JAR stuff works to be sure, but I think this is generally going in the right direction.  What is the story for thin JARs in air-gapped systems?  How do you configure an internal repository to retrieve from?

dsyer:
The repository can be pushed with the app if you need to (launch with `thin.root=.` for example). We'd need to make that easy for users anyway, in case they just want to control it that way.

If you need a remote repository it should work the same way Maven does (i.e. there is a default repository defined in settings.xml). How do people normally control the location of the default repository if it building in an air-gapped system?

Issue #426, opened 2017-05-19
Add IBM JRE as an option to the java-buildpack


dinogun:
Can you please consider adding IBM Java as an option in the java-buildpack. See the [knowledge center](https://www.ibm.com/support/knowledgecenter/SSYKE2/welcome_javasdk_family.html) for more details on IBM Java. We will be submitting a PR shortly for the same. Thank You !

nebhale:
Will do.  Closing in lieu of PR #427.

PR #427, opened 2017-05-19
Add IBM JRE as an option to the java-buildpack

bharathappali:
Please see issue #426 for more details on IBM Java.


lib/java_buildpack/jre/ibmjre.rb
```diff
@@ -0,0 +1,53 @@
...
+
+      # (see JavaBuildpack::Component::ModularComponent#command)
+      def command
+        # no command need to be executed
+      end
+
+      # (see JavaBuildpack::Component::ModularComponent#sub_components)
+      def sub_components(context)
+        [
+          IbmjreLike.new(sub_configuration_context(context, 'jre')
```
nebhale: Unless there is a plan for more than one IBM JRE-like implementation, there's no reason to name this "-like".

lib/java_buildpack/jre/ibmjre_like.rb
```diff
@@ -0,0 +1,174 @@
...
+require 'fileutils'
+require 'java_buildpack/component/versioned_dependency_component'
+require 'java_buildpack/jre'
+require 'java_buildpack/util/tokenized_version'
+
+module JavaBuildpack
+  module Jre
+
+    # Encapsulates the detect, compile, and release functionality for selecting a JRE.
+    class IbmjreLike < JavaBuildpack::Component::VersionedDependencyComponent
```
nebhale: Unless there is a plan for more than one IBM JRE-like implementation, there's no reason to name this "-like".

lib/java_buildpack/jre/ibmjre_like.rb
```diff
@@ -0,0 +1,174 @@
+# Encoding: utf-8
```
nebhale: No need for `Encoding` since we do not support Ruby 1.9.

lib/java_buildpack/jre/ibmjre_like.rb
```diff
@@ -0,0 +1,174 @@
...
+require 'java_buildpack/util/tokenized_version'
+
+module JavaBuildpack
+  module Jre
+
+    # Encapsulates the detect, compile, and release functionality for selecting a JRE.
+    class IbmjreLike < JavaBuildpack::Component::VersionedDependencyComponent
+
+      # constant HEAP_RATIO is assigned the usual heap ratio required
+      HEAP_RATIO = 0.75
```
nebhale: Should be a private constant

spec/java_buildpack/component/base_component_spec.rb
```diff
@@ -1,3 +1,5 @@
+# Encoding: utf-8
```
nebhale: No need for `Encoding` since we do not support Ruby 1.9.

spec/java_buildpack/component/base_component_spec.rb
```diff
@@ -51,6 +53,13 @@
     expect(stdout.string).to match(/Downloading Stub Base Component #{version} from #{uri}/)
   end
 
+  it 'downloads and installs InstallAnywhere (tm) BIN file in the sandbox',
```
nebhale: This functionality is IBM JRE specific and therefore shouldn't be tested as part of this class.

spec/java_buildpack/component/versioned_dependency_component_spec.rb
```diff
@@ -1,3 +1,5 @@
+# Encoding: utf-8
```
nebhale: No need for Encoding since we do not support Ruby 1.9.

spec/java_buildpack/component/versioned_dependency_component_spec.rb
```diff
@@ -48,6 +50,13 @@
       expect(component.detect).to eq("stub-versioned-dependency-component=#{version}")
     end
 
+    it 'downloads and installs InstallAnywhere (tm) BIN file in the sandbox',
```
nebhale: This functionality is IBM JRE specific and therefore shouldn't be tested as part of this class.

spec/java_buildpack/jre/ibmjre_like_spec.rb
```diff
@@ -0,0 +1,59 @@
+# Encoding: utf-8
```
nebhale: No need for `Encoding` since we do not support Ruby 1.9.

spec/java_buildpack/jre/ibmjre_like_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require 'spec_helper'
+require 'component_helper'
+require 'java_buildpack/component/mutable_java_home'
+require 'java_buildpack/jre/ibmjre_like'
+
+describe JavaBuildpack::Jre::IbmjreLike do
```
nebhale: Please follow the formatting style of the other tests.  This means whitespace between specs, etc.

nebhale:
@bharathappali I'm quite interested in getting this into the build pack quickly.  Is there anything specific holding up an updated PR?

bharathappali:
@nebhale Sorry for the delay. Thank you for the prompt review. All comments have been addressed. Please review the latest changes

dinogun:
@nebhale Thank you for super quick response and reviews. Regarding your question on the jvmkill agent. We found that IBM Java differs slightly from OpenJDK in the implementation of the JVMTI spec. We are working to fix those and in the meanwhile @bharathappali has forked the jvmkill and made the changes for it to be able to work with IBM Java. Once the IBM Java implementation is fixed, we will revert to using the standard jvmkill agent.

config/ibm_jre.yml
```diff
@@ -0,0 +1,25 @@
...
+
+# Configuration for JRE repository
+---
+jre:
+  version: 1.8.+
+  repository_root: "https://public.dhe.ibm.com/ibmdl/export/pub/systems/cloud/runtimes/java/meta/jre/linux/x86_64/"
+  heap_ratio: 0.75
+jvmkill_agent:
+  version: 1.+
+  repository_root: "https://raw.githubusercontent.com/bharathappali/jvmkill/jvmkill-ibmagent/killagent/"
```
nebhale: This needs to be hosted in a bit more permanent of a place than just a GitHub repository.  Because that repository can be deleted at any moment, breaking existing users, this should be hosted somewhere a little more canonical.  I'd prefer it it was hosted at or near the other repository, i.e. some place with an IBM domain.

I do understand that it's only temporary, but there will be users that will see this temporary dependency.

lib/java_buildpack/jre/ibm_jre_initializer.rb
```diff
@@ -0,0 +1,228 @@
...
+        @droplet.java_opts.concat mem_opts
+        @droplet.java_opts.concat tls_opts
+        @droplet.java_opts << '-Xshareclasses:none'
+      end
+
+      private
+
+      # constant HEAP_RATIO is the ratio of memory assigned to the heap
+      # as against the container total and is set using -Xmx.
+      HEAP_RATIO = 0.75
```
nebhale: These two should be private constants.  Ruby sadly, doesn't make constants defined after `private`, private.  You'll need an invocation, after the definitions of something like `private_constant :HEAP_RATIO, KILO`, even though they follow the `private` declaration.

lib/java_buildpack/jre/ibm_jre_initializer.rb
```diff
@@ -0,0 +1,228 @@
...
+        @droplet.copy_resources
+      end
+
+      # (see JavaBuildpack::Component::BaseComponent#release)
+      def release
+        @droplet
+          .java_opts
+          .add_system_property('java.io.tmpdir', '$TMPDIR')
+        @droplet.java_opts << '-Xtune:virtualized'
+        @droplet.java_opts.concat mem_opts
```
nebhale: I just want to confirm that you're aware that changes to the memory in a container (i.e. `cf scale <APP> -m <VALUE>`) will require **restaging** the application to observe the changes.  With the OpenJDK-like memory calculator, changes are reflected on a restart.  Are you OK with this deficiency?

lib/java_buildpack/jre/ibm_jre_initializer.rb
```diff
@@ -0,0 +1,228 @@
...
+        end
+        mopts
+      end
+
+      # Returns the heap_ratio attribute in config file (if specified) or the HEAP_RATIO constant value
+      def heap_ratio
+        @configuration['heap_ratio'] || HEAP_RATIO
+      end
+
+      def tls_opts
```
nebhale: I'm a bit skeptical of this setting being the default.  We're seeing more and more customers attempting to lock down the versions of TLS supported (all the way down to locking it to TLS v1.2 _only_).  At the very least, forcing a modification of the defaults seems a bit out of bounds.  A user can always `cf set-env <APP> JAVA_OPTS '-Dcom.ibm.jsse2.overrideDefaultTLS=true'` if they would prefer this behavior.

spec/java_buildpack/jre/ibm_jre_initializer_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require 'spec_helper'
+require 'component_helper'
+require 'java_buildpack/component/mutable_java_home'
+require 'java_buildpack/jre/ibm_jre_initializer'
+
+describe JavaBuildpack::Jre::IbmJreInitializer do
```
nebhale: Can you please add a bit more whitespace to this file to better match [other examples](https://github.com/cloudfoundry/java-buildpack/blob/master/spec/java_buildpack/jre/open_jdk_like_jre_spec.rb#L21)?

spec/java_buildpack/jre/ibm_jre_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+# limitations under the License.
+
+require 'spec_helper'
+require 'component_helper'
+require 'fileutils'
+require 'java_buildpack/component/mutable_java_home'
+require 'java_buildpack/jre/ibm_jre_initializer'
+require 'java_buildpack/jre/ibm_jre'
+
+describe JavaBuildpack::Jre::IbmJRE do
```
nebhale: Can you please add a bit more whitespace to this file to better match [other examples](https://github.com/cloudfoundry/java-buildpack/blob/master/spec/java_buildpack/jre/open_jdk_like_spec.rb#L25)?

config/ibm_jre.yml
dinogun: Hi Ben, I understand your concern about the GitHub repo being temporary, however it may take us quite a while to get auth for pushing files to public.dhe.ibm.com. Do you think using our official github repo (https://github.com/ibmruntimes) is any better ? We currently have all our Dockerfiles for ibmjava over there and it isn't going away any time soon.

lib/java_buildpack/jre/ibm_jre_initializer.rb
dinogun: Yes, we will update this to match OpenJDK in a subsequent patch.

lib/java_buildpack/jre/ibm_jre_initializer.rb
dinogun: We will remove this option and add a recommendation in the readme instead

config/ibm_jre.yml
nebhale: Yeah, that I could handle.  I'd just like IBM's name somewhere in it if it disappears.  It's all about having a large company to blame 😄

lib/java_buildpack/jre/ibm_jre_initializer.rb
nebhale: Understood.

nebhale:
@bharathappali @dinogun If you can get the final changes in on Tuesday, it'll go out in a release that afternoon.  If not, it'll make the next release, a couple of weeks down the line.

bharathappali:
@BenHale We have now updated the PR as per your comments. The jvmkill agent is now hosted at https://github.com/ibmruntimes/jvmkill. Also the recent security provider changes necessitated a change for IbmJRE as well. It is now using OpenJdkLikeSecurityProviders for the same. Please review, Thank you

nebhale:
This is mostly ready go except that your [repository index](https://public.dhe.ibm.com/ibmdl/export/pub/systems/cloud/runtimes/java/meta/jre/linux/x86_64/index.yml) isn't valid.  The only valid repository index format is described [here](https://github.com/cloudfoundry/java-buildpack/blob/master/docs/extending-repositories.md#repository-structure).  Once you update that index, I'll merge this in.

dinogun:
Hi Ben, Unfortunately it is going to be very difficult to change the index.yml format. Changing the format of the file will break a lot of users including the liberty buildpack. Even though the format is not matching the spec as you have pointed out, the current parser seems to be working with the index.yml. Do you think having a custom parser will help ? Not sure how else to resolve this issue.

nebhale:
The buildpack will not knowingly merge functionality that expects non-standard behavior.  It doesn't actually work today and the example is that [this branch](https://github.com/cloudfoundry/java-buildpack/tree/427-ibm-jre) (the merge candidate) do not run successfully now that the code expects a standard repository.  Effectively, that index works in unspecified area of the repository spec and any change to that repository spec in the future could break this functionality.

And to be clear, this requirement doesn't have to be a breaking change for Liberty Buildpack users.  For example [this](https://gist.githubusercontent.com/nebhale/ddc777ef2dda218ecdab2946e020536e/raw/17c7cef370071b82439128c1303e66368fc9a975/index.yml) is now a perfectly legal index.  If you updated your `repository_root` to point at `https://gist.githubusercontent.com/nebhale/ddc777ef2dda218ecdab2946e020536e/raw/17c7cef370071b82439128c1303e66368fc9a975`, everything would just work.  The `index.yml` is intentionally decoupled from the locations of the files, so that integrations can have their binaries exposed wherever, structured however, and named whatever, and the `index.yml` acts as the indirection that makes those existing releases, Java Buildpack compliant.

dinogun:
Hi Ben, it looks like we need to maintain a separate index.yml that conforms to the cf java spec. We are  finalizing the right location and generating it as part of our release process. Will update the PR in a day or two. Thank you for the patience !

nebhale:
No problem at all.  I'm quite looking forward to getting this contribution merged when you're ready.

nebhale:
@dinogun Any news on the index?

dinogun:
Sorry for the late reply, we'll submit it today. Thanks

dinogun:
Thanks Ben !

PR #431, opened 2017-05-24
Add support for binary proxies (like Artifactory/Nexus)

z4ce:
This is designed to allow enterprises using artifactory
and nexus type mirroring solutions to use the online JBP
without having to have offline processing jobs to modify
the index.yml files

Syntax is add a stanza like
```
binary_proxies:
- from: https://java-buildpack.cloudfoundry.org
  to: http://nexus:8081/repository/pcf-javabuildpack
```
to the `config/repository.yml`

### DIFF
--- a/docs/extending-repositories.md
+++ b/docs/extending-repositories.md
@@ -73,6 +73,15 @@ Access to repositories may be affected by the existence of network proxies.  In 
 ```bash
 cf set-env <APP_NAME> http_proxy http://username:password@host:port
 ```
+## Mirroring proxies
+In some cases you may want to mirror https://java-buildpack.cloudfoundry.org or another
+binary repository using a repository manager like artifactory or nexus without modifying the `index.yml` files under the repository. To do so
+add the following to [`config/repository.yml`][] replacing nexus with your mirror
+```
+binary_proxies:
+- from: https://java-buildpack.cloudfoundry.org
+  to: http://nexus:8081/repository/pcf-javabuildpack
+```
 
 ## Version Syntax and Ordering
 Versions are composed of major, minor, micro, and optional qualifier parts (`<major>.<minor>.<micro>[_<qualifier>]`).  The major, minor, and micro parts must be numeric.  The qualifier part is composed of letters, digits, and hyphens.  The lexical ordering of the qualifier is:
@@ -96,4 +105,3 @@ In addition to declaring a specific versions to use, you can also specify a boun
 [`JavaBuildpack::Repository::ConfiguredItem`]: ../lib/java_buildpack/repository/configured_item.rb
 [Configuration and Extension]: ../README.md#configuration-and-extension
 [example]: http://download.pivotal.io.s3.amazonaws.com/openjdk/trusty/x86_64/index.yml
-
--- a/lib/java_buildpack/component/base_component.rb
+++ b/lib/java_buildpack/component/base_component.rb
@@ -20,7 +20,7 @@ require 'java_buildpack/util/format_duration'
 require 'java_buildpack/util/shell'
 require 'java_buildpack/util/space_case'
 require 'java_buildpack/util/sanitizer'
-
+require 'java_buildpack/util/binary_proxy'
 module JavaBuildpack
   module Component
 
@@ -85,6 +85,7 @@ module JavaBuildpack
       # @param [String] name an optional name for the download.  Defaults to +@component_name+.
       # @return [Void]
       def download(version, uri, name = @component_name)
+        uri = JavaBuildpack::Util::BinaryProxy.proxy_for(uri)
         download_start_time = Time.now
         print "-----> Downloading #{name} #{version} from #{uri.sanitize_uri} "
 
--- a/lib/java_buildpack/util/binary_proxy.rb
+++ b/lib/java_buildpack/util/binary_proxy.rb
@@ -1,0 +1,51 @@
+# Cloud Foundry Java Buildpack
+# Copyright 2013-2017 the original author or authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require 'pathname'
+require 'java_buildpack/util'
+require 'java_buildpack/util/configuration_utils'
+require 'java_buildpack/logging/logger_factory'
+require 'shellwords'
+require 'yaml'
+
+module JavaBuildpack
+  module Util
+
+    # Utility for loading configuration
+    class BinaryProxy
+
+      private_class_method :new
+
+      class << self
+
+        # Checks to see if a proxy exists for a URI. If it does, replace it
+        # with the proxy uri.
+        #
+        # @param [String] URI to check for proxies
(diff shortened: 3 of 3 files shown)


z4ce:
I made the `pivotal-cf` org visible so now the bot should be happy if you close/reopen.

psa2211:
We are a large PCF customer and could really use this pull request to support our use cases for the java-buildpack.

nebhale:
The idea of this is reasonable, but it's handled in the wrong location.  The fact that configuration is done in `repository.yml` is indicative of the fact that it should all be handled in the `RespositoryIndex` or thereabouts, rather than at the component level.  The fact that a mapping has happened, should never escape the repository abstraction.  In addition, as it exists today, that mapping is only ever applied to the contents of the index and is not applied to the location of the index.  Moving the functionality into `RepositoryIndex` would allow it to be applied to the entire repository URI resolution, both of indexes and artifacts.

nebhale:
@psa2211 Why is the [repository replication functionality](https://github.com/cloudfoundry/java-buildpack-dependency-builder#replicating-repository) insufficient for you use-cases?  Why isn't it possible to transform the indexes as they are replicated into your local system?

psa2211:
We don't have an S3 compatible store to use the replication script. Even if we did, it is a lot more moving pieces (and firewall rules) and can and will break than just using a pass-through proxy like Nexus.

nebhale:
@psa2211 You don't need to have an S3 compatible store to replicate into.  The replicate script downloads onto any filesystem.  In addition, if you're using a pass-through proxy like Nexus, shouldn't that proxy every single URI, including all the original ones?

psa2211:
Our only approve mechanism for providing repositories is through nexus. It does proxy every URI under a new URI, but it doesn't rewrite the URIs in the index files."

z4ce:
@nebhale if you think this is implemented in the wrong spot, I'm open to other ideas.  This patch I believe handles it at the global level, every object retrieved by the buildpack that references java-buildpack.cloudfoundry.org, for example, will get translated to the new URL.

nebhale:
@psa2211 If Nexus proxying every URI automatically, then doesn't that mean that when a URI is read out of an `index.yml` and a subsequent request is made, that Nexus will also remap the that URI from the `index.yml` as well?

psa2211:
No, Nexus proxies the repo at a new URL, as does artifactory. For example, the repo would be proxied locally within our company at https://nexus.mycorp.com/repositories/java-buildpack

nebhale:
@psa2211 and how is that repository populated?

kelapure:
This PR allows customers to reuse the Java Online Buildpack without having to standup their own repository with the requisite offline dependencies. It allows developers to start using the latest secure JBP without waiting for an offline release.

psa2211:
The repository is configured as a Generic Proxy Repository. It is populated as a pass-through to the official jbp repo. The details for configuring are here:
https://books.sonatype.com/nexus-book/reference/confignx-sect-manage-repo.html

nebhale:
@psa2211 As you’ve described it to me, the generic proxy will map all URIs, whether they come from the buildpack or the `index.yml`s downloaded by the buildpack.  Given that, I’m not sure why you’d need to map _any_ URI anywhere.

nebhale:
@kelapure As described so far, the generic proxy functionality doesn’t appear to require any changes to either the buildpack or even the `index.yml`s.  From the description it would proxy and cache the request for `https://java-buildpack.cloudfoundry.org/openjdk-jdk/trusty/x86_64/index.yml` and then subsequently proxy and cache the request for `https://java-buildpack.cloudfoundry.org/openjdk-jdk/trusty/x86_64/openjdk-1.8.0_141.tar.gz`.

If they wanted to pre-populate a repository internally (not use a proxy) then the [`replicate.sh` script](https://github.com/cloudfoundry/java-buildpack-dependency-builder#replicating-repository) already does the appropriate [syncing](https://github.com/cloudfoundry/java-buildpack-dependency-builder/blob/master/replicate.sh#L17) and [remapping](https://github.com/cloudfoundry/java-buildpack-dependency-builder/blob/master/replicate.sh#L46).

I still haven’t seen a compelling argument as to why running that script, which will both replicate dependencies and remap the `index.yml` files isn’t sufficient.

psa2211:
It maps all URLs at a NEW URLs. For example:

https://java-buildpack.cloudfoundry.org/openjdk/trusty/x86_64/index.yml
becomes
https://nexus.corp.com/repositories/java-buidpack/openjdk/trusty/x86_64/index.yml
and
https://java-buildpack.cloudfoundry.org/openjdk/trusty/x86_64/openjdk-1.8.0_77.tar.gz
becomes
https://nexus.corp.com/repositories/java-buidpack/openjdk/trusty/x86_64/openjdk-1.8.0_77.tar.gz
So if you configure the javabuild pack to use a Nexus mirrored repository it will read the index.yml (from the nexus repo) and then attempt to access the internet (because the index.yml does not use the repo) which will fail in environments without direct internet access

z4ce:
Setting up a generic proxy (like Nexus/Artifactory) does require this PR. As for why this PR is necessary vs the replicate script. This is how a conversation goes if you want to use the online JBP:

Do you have internet access from your PCF foundation? No.

Can you provision a box with a web server, network access to each foundation, and access to the internet?  Maybe, with weeks to months of discussions with security and network change tickets. (a long, time consuming version of No most of the time)

Do you have Nexus or artifactory? Yes. Oh yeah, and its already network available to each PCF foundation for other purposes (nodejs, pypi, etc). And security has already blessed it and scans it regularly for vulnerabilities. 

And the latter solution has no moving parts. It's just configuration with nexus and artifactory. No scheduled shell scripts to monitor, don't have to worry about backups (as nexus/artifactory is already handled), firewall rules already in place, etc. I think it is far to say that lack of nexus/artifactory support is the single biggest impediment keeping organizations using the offline buildpack.

nebhale:
@psa2211 I don't quite understand this statement:

> So if you configure the javabuild pack to use a Nexus mirrored repository it will read the index.yml (from the nexus repo) and then attempt to access the internet (because the index.yml does not use the repo) which will fail in environments without direct internet access

Why doesn't the `index.yml` use the repo?  As you said, the proxy will map `https://java-buildpack.cloudfoundry.org/openjdk/trusty/x86_64/openjdk-1.8.0_77.tar.gz` to `https://nexus.corp.com/repositories/java-buidpack/openjdk/trusty/x86_64/openjdk-1.8.0_77.tar.gz` automatically.  So when the buildpack attempts the request for the `java-buildpack.cloudfoundry.org` URI, the proxy will remap it to a URI it can service.

z4ce:
@nebhale the `index.yml` comes from the repo, but still has the old URLs which specifies fully qualified URLs. When the javabuild pack gets the URL, it attempts to access it directly as specified and not the new URL. The `index.yml` file would need to be modified to replace `https://java-buildpack.cloudfoundry.org` with `https://nexus.corp.com/repositories/java-buidpack` Artifactory/Nexus do not act like a SOCKS proxy, it proxies at a new URL. 

The steps to failure look like this:
- Configure artifactory/nexus to proxy java-buildpack.cloudfoundry.org
- Configure java buildpack `repository.yml` or environment variable to set  `default_repository_root: https://nexus.corp.com/repositories/java-buidpack`
- Push an app using this config
- Build pack runs, and downloads the `index.yml` from the JDK
- The `index.yml` specifies a URL on the internet and the download of the JDK fails

psa2211:
What @z4ce said is exactly our problem. We change default_repository_root to point at our Nexus JBP mirror. The JBP connects out to Nexus and gets a index.yml but then fails when trying to download from the internet.

nebhale:
The next generation of buildpacks, [Cloud Native Buildpacks](https://buildpacks.io) no [longer uses repositories and variable dependency resolution](https://github.com/cloudfoundry/openjdk-buildpack/blob/master/buildpack.toml#L23).  As such, we won't be putting any effort into the current buildpack to support this use-case.

PR #409, opened 2017-03-24
Added FusionReactor monitor support

Umaaz:
I have added support for our FusionReactor APM Java monitoring tool.

## Usage
Create a new service called FusionReactor that defines a password and license key.
e.g.
```
cf cups fusionreactor -p "{\"password\":\"admin\",\"license\":\"........\"}"
```
Then add the 'fusionreactor' service to the application.

PR #446, opened 2017-06-12
Added Contrast Security framework support

donniepropst:
This commit adds support for using Contrast Security's Java agent. 
A bound service will be used to create a configuration file for the agent, as well as set some Java system properties.

### DIFF
--- a/lib/java_buildpack/framework/contrast_security_agent.rb
+++ b/lib/java_buildpack/framework/contrast_security_agent.rb
@@ -1,0 +1,114 @@
+# Encoding: utf-8
+
+# Cloud Foundry Java Buildpack
+# Copyright 2013-2017 the original author or authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require 'fileutils'
+require 'open-uri'
+require 'rexml/document'
+require 'java_buildpack/component/versioned_dependency_component'
+require 'java_buildpack/framework'
+require 'base64'
+
+module JavaBuildpack
+  module Framework
+
+    # Encapsulates the functionality for running the Contrast Security Agent support.
+    class ContrastSecurityAgent < JavaBuildpack::Component::VersionedDependencyComponent
+
+      # (see JavaBuildpack::Component::BaseComponent#compile)
+      def compile
+        load_credentials
+        FileUtils.mkdir_p(@droplet.sandbox)
+        @dir = @droplet.sandbox
+        download_jar(boot_class_name)
+        build_contrast_configuration
+        @droplet.copy_resources
+      end
+
+      # (see JavaBuildpack::Component::BaseComponent#release)
+      def release
+        app_name = @application.details['application_name'] || 'ROOT'
+        java_opts = @droplet.java_opts
+        java_opts.add_system_property('contrast.dir', '/tmp')
+        java_opts.add_system_property('contrast.override.appname', app_name)
+        path = java_opts.qualify_path(@droplet.sandbox)
+        java_opts.add_preformatted_options("-javaagent:#{path}/#{boot_class_name}=#{path}/contrast.config")
+      end
+
+      protected
+
+      # (see JavaBuildpack::Component::VersionedDependencyComponent#supports?)
+      def supports?
+        @application.services.one_service?(CONTRAST_FILTER, [TEAMSERVER_URL, USERNAME, API_KEY, SERVICE_KEY])
+      end
+
+      private
+
+      CONTRAST_FILTER = /contrast[-]?security/
+      private_constant :CONTRAST_FILTER
+
+      TEAMSERVER_URL = 'teamserver_url'.freeze
+      USERNAME = 'username'.freeze
+      API_KEY = 'api_key'.freeze
+      SERVICE_KEY = 'service_key'.freeze
+
+      PLUGIN_PACKAGE = 'com.aspectsecurity.contrast.runtime.agent.plugins.'.freeze
+
+      def load_credentials
+        credentials = @application.services.find_service(CONTRAST_FILTER)['credentials']
+        @teamserver_url = credentials[TEAMSERVER_URL]
+        @username = credentials[USERNAME]
+        @api_key = credentials[API_KEY]
+        @service_key = credentials[SERVICE_KEY]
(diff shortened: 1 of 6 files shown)


nebhale:
@donniepropst I'll start a review on this today, but please get started on the CLA so that we merge as soon as the technical changes are complete.

lib/java_buildpack/framework/contrast_security_agent.rb
```diff
@@ -0,0 +1,114 @@
...
+        download_jar(boot_class_name)
+        build_contrast_configuration
+        @droplet.copy_resources
+      end
+
+      # (see JavaBuildpack::Component::BaseComponent#release)
+      def release
+        app_name = @application.details['application_name'] || 'ROOT'
+        java_opts = @droplet.java_opts
+        java_opts.add_system_property('contrast.dir', '/tmp')
```
nebhale: `/tmp` isn't a reasonable location within the container (there's no guarantee that it is mounted).  You should use `$TMPDIR` instead.

lib/java_buildpack/framework/contrast_security_agent.rb
```diff
@@ -0,0 +1,114 @@
...
+        java_opts.add_system_property('contrast.override.appname', app_name)
+        path = java_opts.qualify_path(@droplet.sandbox)
+        java_opts.add_preformatted_options("-javaagent:#{path}/#{boot_class_name}=#{path}/contrast.config")
```
nebhale: Hmm...  We really need to get an `add_javaagent_with_props` [added](https://github.com/cloudfoundry/java-buildpack/pull/409/files#diff-7b66b1ae901a0fdd479e807c831c9d9b).  Not your responsibility.

lib/java_buildpack/framework/contrast_security_agent.rb
```diff
@@ -0,0 +1,114 @@
...
+      end
+
+      protected
+
+      # (see JavaBuildpack::Component::VersionedDependencyComponent#supports?)
+      def supports?
+        @application.services.one_service?(CONTRAST_FILTER, [TEAMSERVER_URL, USERNAME, API_KEY, SERVICE_KEY])
```
nebhale: This is a bit unclear, but putting these keys in an array means "or".  If you want "and" (and I I'm pretty sure you do), it should just be comma delimited.  `@application.services.one_service?(CONTRAST_FILTER, TEAMSERVER_URL, USERNAME, API_KEY, SERVICE_KEY)`.

lib/java_buildpack/framework/contrast_security_agent.rb
```diff
@@ -0,0 +1,114 @@
...
+      end
+
+      private
+
+      CONTRAST_FILTER = /contrast[-]?security/
+      private_constant :CONTRAST_FILTER
```
nebhale: All of these constants should be alphabetized and added to `private_constants`.

lib/java_buildpack/framework/contrast_security_agent.rb
```diff
@@ -0,0 +1,114 @@
...
+      protected
+
+      # (see JavaBuildpack::Component::VersionedDependencyComponent#supports?)
+      def supports?
+        @application.services.one_service?(CONTRAST_FILTER, [TEAMSERVER_URL, USERNAME, API_KEY, SERVICE_KEY])
+      end
+
+      private
+
+      CONTRAST_FILTER = /contrast[-]?security/
```
nebhale: Since this the first time we've ever integrated, you should pick a single, canonical, representation that you'd like to search for and that should become the truth.

lib/java_buildpack/framework/contrast_security_agent.rb
```diff
@@ -0,0 +1,114 @@
...
+        @api_key = credentials[API_KEY]
+        @service_key = credentials[SERVICE_KEY]
+      end
+
+      def boot_class_name
+        version = @version.to_s.split('_')[0]
+        "contrast-engine-#{version}.jar"
+      end
+
+      def build_contrast_configuration
```
nebhale: I'll leave this one to you to decide, but making a real live XML document might not be the easiest or clearest way to do this.  Another option is to use extended literals and just write clear text that you know to be valid XML.  [An example](https://github.com/cloudfoundry/java-buildpack/blob/master/lib/java_buildpack/framework/luna_security_provider.rb#L146-L163).  It's up to you though, and if you think this is easier, I'm happy to leave it as-is.

donniepropst:
@nebhale thanks for the quick review, I'm hoping to work on the fixes you mentioned later today or early tomorrow, and will touch base with product/legal teams in regard to the CLA.

donniepropst:
@nebhale I believe I resolved all your comments with the exception of 1.) How I built the xml contrast.config file, I would prefer to keep this the same. and 2.) the renaming of the jar.

In our repo the jar is packaged with a META-INF/MANIFEST.MF that specifies the jar name as: 'contrast-engine-#.#.#.jar'.  In SaaS (or for customers hosting our platform internally) our server will  rename the jar and rewrite the manifest file to specify the name as just 'contrast.jar'.  Although we may be able to make a change to our process to match the name and version convention that you specified I don't believe that is something we will be able to do on the short term.

I believe the CLA should be submitted today if it wasn't yesterday.

nebhale:
@donniepropst Just trying to get a better understanding of the JAR naming requirement.  Why does the agent care what the name of the JAR is?  Is it used for more than just an opaque file path passed to `javaagent`?

donniepropst:
@nebhale After talking with one of our Java agent engineers they said the name has to match because the JVM inspects the manifest file and the Boot-Class-Path value gets added to the classpath at boot, so if the jar name doesn't match the information specified in the manifest we see NoClassDefFoundExceptions.  

This was a requirement I didn't realize when I was working on the buildpack and calling download_jar without any parameters.  The jar would get downloaded, but would have a bunch of class not found exceptions from our agent upon startup because the name of the file as stored in our repo is Java-Agent-#.#.#.#.jar.  

The unzipped manifest.mf file:
Manifest-Version: 1.0
Archiver-Version: Plexus Archiver
Created-By: Apache Maven
Built-By: jenkins
Build-Jdk: 1.7.0_80
Main-Class: com.contrastsecurity.agent.ContrastEngine
Agent-Class: com.contrastsecurity.agent.ContrastAgent
Boot-Class-Path: contrast-engine-3.4.0.jar
Build-Time: 20170525.2358
Can-Redefine-Classes: true
Can-Retransform-Classes: true
Can-Set-Native-Method-Prefix: true
Implementation-Build: 3.4.0.639
Premain-Class: com.contrastsecurity.agent.ContrastAgent

nebhale:
OK, understood.  We'll just wait on the CLA and then everything should be good to go.

donniepropst:
Checked in with our product team and it looks like the CLA was accepted last Wednesday so we should be good

nebhale:
@donniepropst When you get a moment can you please ensure that you've got the proper GitHub organization publicized?

donniepropst:
@nebhale that was probably the issue.  It is public now.

nebhale:
Looks good.  Closing as merged now.

PR #457, opened 2017-07-04
Make it possible to disable auto-activation of JRebel

dmitri-gb:


nebhale:
What's the need for this?  Wouldn't you just not package your application without `rebel-remote.xml` if you didn't want that behavior?

dmitri-gb:
@nebhale Yes, one could also just leave out the `xml` file. However, we've had clients who keep this file in their source code repository and deploy to multiple environments (dev, staging, production). They have expressed the wish to be able to disable JRebel through other means (env vars etc).

PR #479, opened 2017-08-14
Add "skiperror" flag for Dynatrace One Agent

PatrikSteuer:
Add `skiperrors` flag to user-provided service for dynatrace one agent.
This flag provides the possibility of a robust startup if the download
of the one agent fails due to any circumstances.

Possible values for skiperrors are:
- false (default) will abbort the staging process
- true will continue staging with a deactivated dynatrace service

### DIFF
--- a/docs/framework-dynatrace_one_agent.md
+++ b/docs/framework-dynatrace_one_agent.md
@@ -29,6 +29,7 @@ The credential payload of the service may contain the following entries:
 | `apitoken` | The token for integrating your Dynatrace environment with Cloud Foundry. You can find it in the deploy Dynatrace section within your environment. The `apitoken` replaces deprecated ~~`tenanttoken`~~ option.
 | `apiurl` | (Optional) The base URL of the Dynatrace API. If you are using Dynatrace Managed you will need to set this property to `https://<your-managed-server-url>/e/<environmentId>/api`. If you are using Dynatrace SaaS you don't need to set this property.
 | `endpoint`  | (Deprecated) The Dynatrace connection endpoint the agent connects to. Please use `apiurl` in combination with `apitoken` for Dynatrace Managed.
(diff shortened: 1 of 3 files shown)


nebhale:
@PatrikSteuer We don’t skip errors on staging failures like this.  We consider that running without the expected agents attached is just as bad as not running at all as it gives you a false sense of confidence that things are working as you expect.  I do not expect to merge this PR, but am open to being convinced.

michele-mancioppi:
Hi @nebhale, here the difference between Dynatrace Managaed and most other APM vendors is that their agent instrumentation is always downloaded from the cluster where the data is sent, instead of "separately" from a simple download endpoint. It can be the case that, during updates of the cluster or any other reason of downtime or network disruption, this prevents CF apps from staging. I think it is a rather unique situation as far as APM vendors are concerned.

The only solution we know for the case is to remove the binding, but that is pretty tough to automate (grepping staging lots in case of push fail).

The best solution would be for Dynatrace to give the option to download non-configured agent instrumentation packages independently from the cluster, but until then, I think this option is at least worth discussing about.

nebhale:
After discussing with @michele-mancioppi I'm going to close this out.  My reasoning is as follows:

I consider silently not binding to an APM problematic. For example, what if your alerting is done by Dynatrace? You push via CI and your application doesn't bind to the APM.  You're basically unprotected at that point, and you don't realize it because this non-binding was done in an automated fashion and you don't realize it.  Combine that with the fact that you don't know when the next restate is coming, so you don't know when it will sort itself out.

I also don't think that this is a Dynatrace-specific issue (related to their agent download system).  Internal repositories going down for maintenance would cause the same effect.

I'm happy to work with Dynatrace on the root cause of this (that their agent download is unavailable too often compared to a static repository) to come up with a robust system that still makes it apparent when your application is unprotected.

michele-mancioppi:
For the record, the availability of the Dynatrace SaaS is pretty awesome :-)

We just want to avoid issues during sporadic update procedures in on-prem installations.

PR #480, opened 2017-08-16
Prepare Contrast framework for boot class path change

donniepropst:
With our next on premise release(3.4.3) the Contrast Java agent's boot class path will be
`java-agent-#{major.minor.patch}.jar`

In order to support previous versions 3.4.0 - 3.4.2 we need to be able to continue using: `contrast-engine-#{major.minor.patch}.jar` and for new versions(3.4.3+) use `java-agent-#{major.minor.patch}.jar`

### DIFF
--- a/lib/java_buildpack/framework/contrast_security_agent.rb
+++ b/lib/java_buildpack/framework/contrast_security_agent.rb
@@ -39,9 +39,9 @@ module JavaBuildpack
       # (see JavaBuildpack::Component::BaseComponent#release)
       def release
         @droplet.java_opts
-                .add_system_property('contrast.dir', '$TMPDIR')
-                .add_system_property('contrast.override.appname', application_name)
-                .add_preformatted_options("-javaagent:#{qualify_path(@droplet.sandbox + jar_name, @droplet.root)}=" \
+          .add_system_property('contrast.dir', '$TMPDIR')
+          .add_system_property('contrast.override.appname', application_name)
+          .add_preformatted_options("-javaagent:#{qualify_path(@droplet.sandbox + jar_name, @droplet.root)}=" \
                                           "#{qualify_path(contrast_config, @droplet.root)}")
       end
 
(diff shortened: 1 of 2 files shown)


lib/java_buildpack/framework/contrast_security_agent.rb
```diff
@@ -49,7 +49,13 @@ def release
 
       # (see JavaBuildpack::Component::VersionedDependencyComponent#jar_name)
       def jar_name
-        "contrast-engine-#{@version.to_s.split('_')[0]}.jar"
+        boot_version = @version.to_s.split('_')[0]
+        major, minor, patch = boot_version.split('.').map(&:to_i)
```
nebhale: No need to do this split and comparison, the `TokenizedVersion` is comparable.  [Example](https://github.com/cloudfoundry/java-buildpack/blob/3b0fa53b5ad50ded173f73ec95e7c517fb9f60bd/lib/java_buildpack/component/mutable_java_home.rb#L39)

nebhale:
@donniepropst Have another go with the simpler version of that comparison.

lib/java_buildpack/framework/contrast_security_agent.rb
```diff
@@ -49,7 +49,13 @@ def release
 
       # (see JavaBuildpack::Component::VersionedDependencyComponent#jar_name)
       def jar_name
-        "contrast-engine-#{@version.to_s.split('_')[0]}.jar"
+        boot_version = @version.to_s.split('_')[0]
```
nebhale: Depending on how you feel about this, the more idiomatic way to do this same thing would be to do

`boot_version = "#{@version[0]}.#{@version[1]}.#{@version[2]}"`

since the `_<QUALIFIER>` isn't a required part of the version spec.  The first three numeric components are required and guaranteed to exist.

donniepropst:
@nebhale thanks for the feedback, I pushed the requested changes

PR #487, opened 2017-09-12
Change service detection for dynatrace integration

stipx:
Things done:
* Service detection also takes the required credentials into account
* Removed deprecated integration option
* Determine agent loader location via manifest
* Renamed the environment variables from RUXIT_* to DT_*

Why?
The integration using "server", "tentanttoken", etc was deprecated quite some time ago and is now removed in favor of the "environmentid", "apitoken" (and optional "apiurl") integration option.
It is possible to define multiple services containing "dynatrace" (no "ruxit" anymore) in order to be able to attach application tags which are read by the agent (tests include multiple services). The only one service containing "environmentid" and "apitoken" is used for the agent integration. The integration is only done iff one service containing these credentials is present (just like before, but also working in release stage now).
Also the agent properties are now also read from environment variables. Instead of parameters on the java agent path.

I also have already contributed to other CF buildpacks and therefore have signed the CLA.

### DIFF
--- a/lib/java_buildpack/framework/dynatrace_one_agent.rb
+++ b/lib/java_buildpack/framework/dynatrace_one_agent.rb
@@ -30,7 +30,7 @@ module JavaBuildpack
       # @param [Hash] context a collection of utilities used the component
       def initialize(context)
         super(context)
-        @version, @uri = agent_download_url if supports? && supports_apitoken?
+        @version, @uri = agent_download_url if supports?
       end
 
       # (see JavaBuildpack::Component::BaseComponent#compile)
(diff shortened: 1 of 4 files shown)


stipx:
@nebhale I updated the documentation and found two flaws of the changes I did before which I fixed together with the documentation changes.

lib/java_buildpack/framework/dynatrace_one_agent.rb
```diff
@@ -46,83 +46,84 @@ def compile
...
-      ENDPOINT = 'endpoint'.freeze
-
-      private_constant :FILTER, :RUXIT_APPLICATION_ID, :RUXIT_HOST_ID, :SERVER, :TENANT, :TENANTTOKEN, :APITOKEN
-      private_constant :ENVIRONMENTID, :ENDPOINT, :APIURL
+      private_constant :FILTER, :DT_APPLICATION_ID, :DT_HOST_ID
+      private_constant :DT_TENANT, :DT_TENANTTOKEN, :DT_CONNECTION_POINT
+      private_constant :ENVIRONMENTID, :APITOKEN
+
+      def service
+        candidates = @application.services.select do |candidate|
```
nebhale: Why did you replicate the `Services.one_service?` method here?  Do you believe that it works differently than the current implementation?

lib/java_buildpack/framework/dynatrace_one_agent.rb
```diff
@@ -46,83 +46,84 @@ def compile
...
-      def agent_dir
-        @droplet.sandbox + 'agent'
+        candidates.one? ? candidates.first : nil
       end
 
       def agent_path
-        libpath = agent_dir + 'lib64/liboneagentloader.so'
-        libpath = agent_dir + 'lib64/libruxitagentloader.so' unless File.file?(libpath)
-        libpath
+        technologies = JSON.parse(File.read(@droplet.sandbox + 'manifest.json'))['technologies']
```
nebhale: You probably don't want to read this manifest multiple times.  I'll clean up in a polishing commit.

lib/java_buildpack/framework/dynatrace_one_agent.rb
stipx: The one_service implementation only returns true or false if such service is present.
find_service takes the first service it finds, but it does not take the credentials into account. So that is the difference. It searches for one service with the needed credentials and returns it.

stipx:
@nebhale To clarify my comment on the service detection further: The main purpose for this pull request is the changed service detection. This means that multiple "dynatrace-something" services are possible but only one with "environmentid" and "apitoken" set. 
Services.one_service checks if there is only one "dynatrace-something" service and this one service has environmentid and apitoken set.

So basically edf611e reverts the main purpose of this PR.

I completely agree that the manifest reading makes sense to be done only once, but the service detection was changed on purpose. There is nothing replicated here. That's also why the test contains two "dynatrace-*" services to test if the detection works with multiple bound services.

nebhale:
> This means that multiple "dynatrace-something" services are possible but only one with "environmentid" and "apitoken" set.
Services.one_service checks if there is only one "dynatrace-something" service and this one service has environmentid and apitoken set.

Then this is what we should change.  `one_service?` (and the other related methods) should be loosened such that they can select one from multiple candidates rather than expecting a single candidate.  Why don't you open an issue around this and I'll take care of it.

> One addition to the doc changes. I've seen you rearranged the parameter table.

Yep, we always sort the parameter table alphabetically in order to facilitate quickly looking up parameters.  If you find other documentation that isn't that way, we'd consider it a bug to be sorted.

stipx:
I thought it would be a too big change for one integration. 

Thanks for taking care of this. I'll create an issue.

nebhale:
No problem being conservative.  I'll get the change in before the next release.

PR #498, opened 2017-10-05
JAV-00 Added seclore cert

fbideau:


nebhale:
Note that the buildpack already provides unlimited strength encryption since [v3.7.1](https://github.com/cloudfoundry/java-buildpack/releases/tag/v3.7.1) about 18 months ago.

PR #501, opened 2017-10-18
Clarifying the requirements for heap dump generation

jtuchscherer:
I hope this additional detail in the docs can help someone save some time.


jtuchscherer:
I changed the files about the Zulu and Oralce JRE as well (didn't even think about them since I always ever have been using the openjdk).

I amended the change and force pushed to my fork, so that it is only one commit.

PR #520, opened 2017-11-10
Update JAVA_OPTS documentation

youngm:
A stab at updating some JAVA_OPTS documentation for JBP4.

### DIFF
--- a/docs/framework-java_opts.md
+++ b/docs/framework-java_opts.md
@@ -67,44 +67,32 @@ from_environment: false
 java_opts: -Xloggc:$PWD/beacon_gc.log -verbose:gc
 ```
 
-## Memory Settings
-
-The following `JAVA_OPTS` are restricted and will cause the application to fail deployment.
-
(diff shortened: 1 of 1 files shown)


nebhale:
Looks great.  What do you think about removing `-Xmx` specifically.  I cannot think of a single case where setting that would be anything but wrong.

youngm:
We could.  Though I actually set that for an application, as a stop gap, that was having trouble staying within the container memory size while I study the problem.  We set it to a value that we believed should leave plenty of head room for everything else.  It was convenient.

That said I won't have a problem with making it an undocumented feature if you would prefer.

nebhale:
Yeah, let's make it undocumented.  I understand your use case, but I don't want any indication in our docs that it's an appropriate thing for most users to do.  Those who really need it should be able to divine that it's possible for the pattern of the rest.

youngm:
@nebhale done.

PR #525, opened 2017-11-16
Use our docker image

jimmykarily:
The upstream docker image doesn't have the rbenv versions we need. Still more problems to fix but this is a first step.


jimmykarily:
Oops wrong base, reopening

Issue #524, opened 2017-11-15
Retry download of a dependency with back off


stipx:
Hi,

as far as I can see there is a retry mechanism in on downloads of dependencies. The thing is that there seems to be no back off strategy implemented. The retry is just done right away:
https://github.com/cloudfoundry/java-buildpack/blob/9d3bc1968af27bf96deeb1688a8438d5ba95a2f4/lib/java_buildpack/util/cache/download_cache.rb#L330-L336
It might make sense to have some kind of exponential back off there to give the download endpoint some time to recover if there are some serverside problems.

Please share your thoughts on this.

Best
stipx

nebhale:
Seems reasonable.

nebhale:
Any thoughts about the backoff intervals you'd see as helpful in practice?

stipx:
I always find the exponential back off pretty useful. It reduces backpressure but also picks up small hiccups pretty fast.

The function to get the wait interval is:
sleep = min(cap, base + 2 ** attempt)

I usually use a pretty small 'base' like 5 seconds and a 'cap' of around 60 to 90 seconds.

nebhale:
Right, I was wondering if you were thinking of a cap in the single second range or larger.  Let me poll around some others and see what the consensus is and I’ll put something in.

PR #526, opened 2017-11-20
Dynatrace OneAgent integration skiperrors flag

stipx:
Hi,

we got the request to include a "skiperrors" flag from some of our large customers which have the Dynatrace Managed (on premise) solution installed on their servers and use it to monitor their CF applications.

Basically the thing is that they wanted to be able to skip download errors which might occur during upgrades, etc. The main thing here is that we only give the possibility to skip monitoring in case the download fails. It will never be enabled by default. Additionally there are a lot of warnings in to signalize the user that this should only be used if there is no other possibility for the customer and also he does not care about monitoring being disabled for some time. It lies in the customers responsibility to NOT use the option if there are critical use cases based on the monitoring.

We also have the same mechanism in on the nodejs-buildpack and would like to streamline all our integrations to have the same options available.

Best,
stipx

### DIFF
--- a/lib/java_buildpack/framework/dynatrace_one_agent.rb
+++ b/lib/java_buildpack/framework/dynatrace_one_agent.rb
@@ -31,14 +31,24 @@ module JavaBuildpack
       def initialize(context)
         super(context)
         @version, @uri = agent_download_url if supports?
+        @logger        = JavaBuildpack::Logging::LoggerFactory.instance.get_logger DynatraceOneAgent
       end
 
       # (see JavaBuildpack::Component::BaseComponent#compile)
       def compile
-        JavaBuildpack::Util::Cache::InternetAvailability.instance.available(
-          true, 'The Dynatrace One Agent download location is always accessible'
-        ) do
-          download(@version, @uri) { |file| expand file }
+        begin
+          JavaBuildpack::Util::Cache::InternetAvailability.instance.available(
+            true, 'The Dynatrace One Agent download location is always accessible'
+          ) do
+            download(@version, @uri) { |file| expand file }
+          end
+        rescue StandardError => e
+          raise unless skip_errors?
+
+          @logger.error { "Dynatrace OneAgent download failed: #{e}" }
+          @logger.warn { "Agent injection disabled because of #{SKIP_ERRORS} credential is set to true!" }
+          FileUtils.mkdir_p(File.dirname(error_file))
+          File.write(error_file, e.to_s)
         end
 
         @droplet.copy_resources
@@ -46,19 +56,16 @@ module JavaBuildpack
 
       # (see JavaBuildpack::Component::BaseComponent#release)
       def release
-        credentials           = @application.services.find_service(FILTER, APITOKEN, ENVIRONMENTID)['credentials']
-        environment_variables = @droplet.environment_variables
-        manifest              = agent_manifest
+        if File.exist?(error_file)
+          @logger.warn { "Dynatrace OneAgent injection disabled due to download error: #{File.read(error_file)}" }
+          return
+        end
+
+        manifest = agent_manifest
 
         @droplet.java_opts.add_agentpath(agent_path(manifest))
 
-        environment_variables
-          .add_environment_variable(DT_TENANT, credentials[ENVIRONMENTID])
-          .add_environment_variable(DT_TENANTTOKEN, tenanttoken(manifest))
-          .add_environment_variable(DT_CONNECTION_POINT, endpoints(manifest))
-
-        environment_variables.add_environment_variable(DT_APPLICATION_ID, application_id) unless application_id?
-        environment_variables.add_environment_variable(DT_HOST_ID, host_id) unless host_id?
+        dynatrace_environment_variables(manifest)
       end
 
       protected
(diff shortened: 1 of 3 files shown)


nebhale:
@stipx Apologies for the delayed response to this; SpringOne really becomes all consuming.

I've actually had another request for this change (from one of your customers), but I'm very much not a fan of this change.  I absolutely consider an application starting without *expected* APM integration to be a situation so poor that it _should_ be a failure.

Imagine a scenario where a user sets this as part of their manifest (or a buildpack fork).  Since most pushing is done via automated pipelines, a user will never see any indication that their latest application version is not instrumented properly.  Then, once something bad is happening and they head over to Dynatrace to find out what the story is, they realize that they haven't been gathering APM data for the last week.

This situation, that a user expects an integration that silently fails to apply, is something I'm quite against as it reflects poorly on Cloud Foundry.  I strongly feel that any integration, whether it's an APM, an HSM, a database, or anything else should be conservative and fail in the case that the integration cannot be completed successfully.

stipx:
@nebhale no worries. We're all pretty busy at this time of the year ;)

Thanks for implementing #524. This also eases up such service or network interruptions.

I agree with you in most of the parts, but for the first concern that it should be a failure if the expected APM integration is not there: Dynatrace will inform you when a host that was first there is gone. I also see it that way that the user has to set the flag explicitly and then besides expecting the integration to work he also expects it to fail (and only warn about it in the logs and also show the offline host in the monitoring). That's also the point why it is not a thing that would default to be enabled.

If the user chooses to suppress errors in case of issues with the agent download I would say it lies in his responsibility to be aware that there MIGHT be a monitoring data loss.

I always have the opinion that failing is the better option, but I also understand that customers want to at least have the possibility to change the behavior to something that suits their needs.

About your examples with different integrations: Without a database the application would not be possible to run, but without monitoring it would. And as I said, I think leaving this up to the user is the best option.

michele-mancioppi:
I have to second this one. There is a very significant demographics of users that do not believe monitoring to be a functional requirements. And I know from painful experience that there is little to be gained by telling otherwise. Hopefully they'll eventually learn better ;-)

I think a good tradeoff is to allow such a {{skip_errors}} option, but leave as default behavior that the staging will fail unless the agent can be retrieved.

nebhale:
> Without a database the application would not be possible to run, but without monitoring it would.

I think this is the crux of our disagreement really.  I’d argue that a system that expected to be run with monitoring isn’t “running” if that monitoring isn’t there.  It may have started, but a key function of the application (observability and serviceability) is missing.  A very long stretch of an analogy would be a microservice starting and not being able to reach another microservice.  There might be a circuit breaker and fallback to some default behavior, but no one would think that that service was healthy if it could never reach the other end.

I’m interested in the idea that Dynatrace can somehow proactively notify a user if their application monitoring has been interrupted.  Given that unique instances come and go, often dozens of times a day, what heuristics does the platform use when deciding that an instance disappearing is abnormal?

danirabiser:
Hi @nebhale, 
from a product management perspective allowing Dynatrace customers to successfully push their application even when the download of Dynatrace OneAgent fails (for example, because of a network connection error) is cruicial. Please note, by default, an unsuccessful download of Dynatrace OneAgent will result in a deployment error. So `skiperrors` is an opt-in functionality that needs to be explicitly enabled by customers.
Wrt Dynatrace heuristics for deciding that an instance disappearing is abnormal I'd like to point you to a blog post around process group anomaly detection: https://www.dynatrace.com/blog/advanced-process-group-availability-monitoring/
We plan to also provide similar anomaly detection for Cloud Foundry applications.

lib/java_buildpack/framework/dynatrace_one_agent.rb
```diff
@@ -31,34 +31,41 @@ class DynatraceOneAgent < JavaBuildpack::Component::VersionedDependencyComponent
...
 
       # (see JavaBuildpack::Component::BaseComponent#compile)
       def compile
-        JavaBuildpack::Util::Cache::InternetAvailability.instance.available(
-          true, 'The Dynatrace One Agent download location is always accessible'
-        ) do
-          download(@version, @uri) { |file| expand file }
+        begin
+          JavaBuildpack::Util::Cache::InternetAvailability.instance.available(
+            true, 'The Dynatrace One Agent download location is always accessible'
```
nebhale: Ironically, not actually true.

lib/java_buildpack/framework/dynatrace_one_agent.rb
```diff
@@ -31,34 +31,41 @@ class DynatraceOneAgent < JavaBuildpack::Component::VersionedDependencyComponent
...
+        @logger        = JavaBuildpack::Logging::LoggerFactory.instance.get_logger DynatraceOneAgent
       end
 
       # (see JavaBuildpack::Component::BaseComponent#compile)
       def compile
-        JavaBuildpack::Util::Cache::InternetAvailability.instance.available(
-          true, 'The Dynatrace One Agent download location is always accessible'
-        ) do
-          download(@version, @uri) { |file| expand file }
+        begin
```
nebhale: Ruby has a syntax that allows you to embed the `begin/rescue` into the method declaration.  I'll update in a following commit.

lib/java_buildpack/framework/dynatrace_one_agent.rb
```diff
@@ -143,6 +172,10 @@ def host_id?
         @application.environment.key?(DT_HOST_ID)
       end
 
+      def skip_errors?
+        'true'.casecmp(credentials[SKIP_ERRORS] || 'false').zero?
```
nebhale: We've got a `to_b` method that will convert that string to a `boolean` value.  I'll update in a following commit.

nebhale:
I'm going to merge this change, but am using this thread to register a strong disagreement with this choice.  I believe this results in a more dangerous situation for Cloud Foundry users, whether they have to opt in or not (tons of precedent for people copying configuration parameters without understanding what they mean) and makes the Dynatrace integration riskier than any of the other APM integrations.

stipx:
Thanks @nebhale for merging. Also your polishing looks pretty neat. Did not know about the rescue in the method declaration.

Issue #534, opened 2017-12-22
Out of memory when using the Takipi Agent


christophe-f:
Hi,

We have about 40 Java microservices deployed on PWS. Last week, I have enabled OverOps/Takipi agent on a few of these services. All the apps using the agent where crashing (due to Out of memory error) every few minutes/hours. The others apps were fine without it.

About 24 hours ago, I've removed the agent from the crashing apps and everything is fine.

I'm using the buildpack v.4.7

Would be nice to be able to use it without any crashes :) Is there any memory tweak recommendation for this agent?

Thanks,
Christophe

nebhale:
What does the Out of Memory error look like?  Is it a Java OOM (complete with a stack trace and our `jvmkill` output) or is it a container out memory error (`exit 137`)?

christophe-f:
Hi Ben,

This is the error:
`index: 0, reason: CRASHED, exit_description: APP/PROC/WEB: Exited with status 137 (out of memory)`

It's a exit 137

nebhale:
@christophe-f I don't know a whole lot about the OverOps agent, but given the very large number of native libraries is it possible that agent is using a large amount of off-heap memory for itself?  The error message you've provided indicates that the JVM is fine, but something outside of the JVM has requested enough memory that the container's memory limit has been exceeded.

/cc @glyn

christophe-f:
Hey Ben,

Looks like it running a local collector alongside the agent. I will try with a remote collector and let you know the outcome.

christophe-f:
Quick update on the issue. I got everything running fine using the remote collector. The Takipi agent configuration page could be updated from "optional" to "highly recommended" :)

nebhale:
Happy to take a PR for it.

More generally, side cars are problematic for Java applications because we configure the JVM to occupy the _entire_ memory space of the container (value for money and all that).  There's some work being done to make buildpacks more sidecar aware in the future which would help with this, but remote collector is certainly the way to go today.

PR #542, opened 2018-01-21
Overops support improvement

christophe-f:
Add a troubleshooting section related to issue #534 

Loading the Takipi agent after NewRelic or AppDynamics will slow the app startup and could cause a timeout. Now the OverOps agent is loading before any APM.

### DIFF
--- a/config/components.yml
+++ b/config/components.yml
@@ -39,6 +39,7 @@ jres:
 # The JavaOpts is last in order to allow any user-defined JAVA_OPTS to override contributions from earlier frameworks
 frameworks:
   - "JavaBuildpack::Framework::MultiBuildpack"
+  - "JavaBuildpack::Framework::TakipiAgent"
   - "JavaBuildpack::Framework::AppDynamicsAgent"
   - "JavaBuildpack::Framework::AspectjWeaverAgent"
   - "JavaBuildpack::Framework::ClientCertificateMapper"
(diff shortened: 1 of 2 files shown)


config/components.yml
```diff
@@ -39,6 +39,7 @@ jres:
 # The JavaOpts is last in order to allow any user-defined JAVA_OPTS to override contributions from earlier frameworks
 frameworks:
   - "JavaBuildpack::Framework::MultiBuildpack"
+  - "JavaBuildpack::Framework::TakipiAgent"
```
nebhale: Can you please elaborate on why this is necessary?

christophe-f: If the Takipi agent is loaded after the APM agent, it will take longer to start the app on PCF/PWS. Many times it will reach the timeout limit and not be able to deploy/start the app.

nebhale:
@christophe-f Can you please take care of the CLA stuff from [here](https://github.com/cloudfoundry/java-buildpack/pull/542#issuecomment-359255853)?

config/components.yml
nebhale: Doesn't this work in reverse as well though?  Since APM agents are instrumenting a bunch of stuff doesn't that all slow down if OverOps is in place already?  There's a fundamental "slow startup when adding an agent" problem that isn't OverOps-specific that you deal with [by setting `CF_STARTUP_TIMEOUT`](https://docs.cloudfoundry.org/devguide/deploy-apps/large-app-deploy.html#consid_limits).

nebhale:
@christophe-f how goes the [CLA process](https://github.com/cloudfoundry/java-buildpack/pull/542#issuecomment-359255853)?  This has been outstanding for a while and I'd like to get it into the next release.

christophe-f:
@nebhale sorry for the delay, just submitted the CLA.

nebhale:
Thanks.  I'll keep an eye out for it.

PR #546, opened 2018-01-29
add support for Tomcat internalProxies configuration

JamesClonk:
Allows to configure internal proxies for Tomcat.

Signed-off-by: Fabio Berchtold <fabio.berchtold@swisscom.com>

### DIFF
--- a/config/tomcat.yml
+++ b/config/tomcat.yml
@@ -18,11 +18,11 @@
 tomcat:
   version: 8.5.+
   repository_root: "{default.repository.root}/tomcat"
-  context_path: 
+  context_path:
   external_configuration_enabled: false
 external_configuration:
   version: 1.+
-  repository_root: 
+  repository_root:
 lifecycle_support:
   version: 3.+
   repository_root: "{default.repository.root}/tomcat-lifecycle-support"
@@ -42,3 +42,7 @@ redis_store:
 geode_store:
   version: 0.+
   repository_root: https://repo.spring.io/ext-release-local/geode-store
+internal_proxies:
+  version: 0.+
+  repository_root:
+  regex:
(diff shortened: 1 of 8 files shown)


MatthiasWinzeler:
Some context about this PR:

As a CF provider, our Gorouter IPs run in the 100.64 range. Tomcat's `RemoteIPValve` by default only trusts **private IP ranges** for headers like `X-Forwarded-For` and `X-Forwarded-Proto`.

This causes the apps to receive wrong remote adresses (since X-Forwarded-For is not respected) and wrong protocol (since X-Forwarded-Proto is not respected). This breaks Apps which rely on these interfaces, i.e. for redirects.

Also see this discussion upstream: https://bz.apache.org/bugzilla/show_bug.cgi?id=60751

This PR would give us the possibility to configure `RemoteIPValve` with our list of `internalProxies`. For example, we could set `JBP_CONFIG_TOMCAT='{ internal_proxies: { regex: "1.2.3.4" }}'` as `cf staging-environment-variable-group` in our CF installations, which would fix Tomcat's `RemoteIPValve`.

MatthiasWinzeler:
On a second note, it would probably be nicer if we can use another env var to specify this setting than `JBP_CONFIG_TOMCAT `. Since if we set it as a platform operator (with `staging-environment-variable-group`), the user might overwrite this (i.e. when specifying a tomcat version for his app).

Is there a way in the buildpack framework to use another env var?

nebhale:
There hasn't been a strong request this feature beyond its original audience. Given that this is an expansion of configuration for Tomcat and down the slippery slope of the entirety of Tomcat's configuration, I'm not going to merge it in.

PR #553, opened 2018-02-14
Added support for Riverbed Appinternals Agent

fwanggg:
The pull request adds support for Riverbed Appinternals Agent.

- Please guide the process of publishing this java build pack to CF's public github repo.
- Assuming it's easy enough to maintain, We are also interested in hosting artifacts in pivotal cloud foundry's public blobstore. Please guide us as well.
- CLA has been submitted to company's legal to review. I guess we could proceed with the code review process before that's signed.


fwanggg:
@nebhale 

- Can you facilitate that?
Do you mean a PCF/PWS account? If so, i am not sure how to setup one. We've been using pie-27(provided by PCF) and a private PCF cluster(that no one can access outside the company). We are still in development mode for tiles so the official integration on PCF end to end is not working. Would it be possible to narrow down the scope to just test the buildpack integration itself? Locally, we have tested this by using cf cli with _cf cups bla_appinternals_

- what can you tell me about the agent itself? Where can I download it and how is it licensed?
The agent is an APM agent for Java. The javabuildpack sets up -agentpath so the our instrumentation profiler can be hooked up into JVM. The agent enhances the java bytecode at runtime. The artifact itself contains all necessary files/shared objects/jar/classes that are needed at runtime. The agent will not persist any data on local file system. Let me know if you need to know more about the agent, if it's not appropriate to talk publicly, we can chat in the private slack channel(riverbed).

- where can i download it and how is it licensed?
hosting the artifacts is a topic that i wish to discuss with you guys and we are interested in hosting them on PCF's public repo. Currently, the agent artifact is hosted in a temporary location on azure blobstore(we tested the integration locally by pointing root_repository to that temp location). 

Current architecture doesn't require us to authenticate at JBP level(both downloading and data transmitting). We will authenticate/license at bosh Addon level.  

Let me know if you have more questions. And let me know if you need to know the temp location on azure where we stores our agent zip file for testing/other purpose. I can send you via private channel.

nebhale:
What I'm looking for is a Riverbed account.  In order to test that this integration works properly, I'll need to hook up an application and ensure that data is being transferred properly.  Can you facilitate getting an account for the team to test the integration?

As far as agent distribution goes, we prefer if third parties [host their own repositories](https://github.com/cloudfoundry/java-buildpack/blob/master/docs/extending-repositories.md#repository-structure) since they can add to them automatically every time they do a release (a number already do this, but a good example would be [AppDynamics](https://packages.appdynamics.com/java/index.yml).  In the rare case that the third party cannot do this, we would be able to host the artifacts in Pivotal's (**not** the Cloud Foundry Foundation's) repository.  In this case, we'd still need a more-or-less public way of noticing and collecting new releases, where they would then be published into a public location for all users to access.

fwanggg:
Ben, thanks for explanation. I can get an SaaS account for you guys for testing. i don't see a problem with that.
However, does that mean the product has to work end to end to pass your test. Our architecture is perhaps different from other similar solutions, where agent may transmit data to saas server directly. Ours require a middle which would require the tile to be ready(it's not ready at the moment). If tile is not ready, testing environment would pretty hard to setup(i.e. our local setup and pie-27).
With this being said, did i create this PR too early?

nebhale:
@fwanggg I don't have a problem looking over the PR now to get it polished up, but it means that we won't be able to merge it until it's testable.  Just to confirm, there's no scenario where an application sends data directly to the SaaS?

fwanggg:
@nebhale confirmed, application does not send data directly to SaaS. thank you. we are getting close to have the tile ready.

fwanggg:
@nebhale  My company request me to close this pull request for the moment due to some legal process. I will open another pull request once that's settled and we should have a testable solution then i hope. Sorry about the inconvenience.

nebhale:
Understood.

PR #578, opened 2018-04-17
Check JAVA_OPTS for contrast.override.appname before using the settin…

donniepropst:
The `contrast_security_agent` framework was previously setting the 'contrast.override.appname option explicitly which was overriding any value that was provided by the user in the JAVA_OPTS env var.  

This checks the JAVA_OPTS env var for the 'contrast.override.appname' system property and if found it doesn't use the app name provided.

### DIFF
--- a/lib/java_buildpack/framework/contrast_security_agent.rb
+++ b/lib/java_buildpack/framework/contrast_security_agent.rb
@@ -39,9 +39,9 @@ module JavaBuildpack
 
       # (see JavaBuildpack::Component::BaseComponent#release)
       def release
+        @droplet.java_opts.add_system_property('contrast.override.appname', application_name) unless app_name_defined?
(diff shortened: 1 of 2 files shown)


lib/java_buildpack/framework/contrast_security_agent.rb
```diff
@@ -109,6 +109,10 @@ def application_name
         @application.details['application_name'] || 'ROOT'
       end
 
+      def app_name_defined?
+        @droplet.java_opts.as_env_var.include?('contrast.override.appname')
```
nebhale: The more Ruby idiomatic way to have done this was with `java_opts.any? { |java_opt| java_opt =~ 'contrast.override.appname')`.  I'll polish up in a following commit.

PR #599, opened 2018-06-07
Append a variable to the nodename property in the appdynamics_agent.yaml

teja1989:
We are having application running in multiple DC's . We are trying to monitor the APP under common Application name "ABC" . To differentiate the cf instances at node level , we are trying to append Datacenter name to the node name property. I am passing this "DC" variable in the Userprovided services.


nebhale:
Why do you believe that this change should be in the main Java Buildpack for all users?

PR #602, opened 2018-06-18
Update introscope agent process name property

DMevada:
This change tweaks the existing property for the agent process name to make it configurable.
 This property is passed via the service credential payload. Before this change, the user could not specify the process name for the agent, and after this change, the user can optionally specify the process name, or let it default to the application name.

### DIFF
--- a/docs/framework-introscope_agent.md
+++ b/docs/framework-introscope_agent.md
@@ -23,7 +23,7 @@ The credential payload of the service may contain the following entries:
 
 | Name | Description
 | ---- | -----------
-|`agent_process_name`| (Optional) The name that is specified for the agent process. If not specified, default value is the application name.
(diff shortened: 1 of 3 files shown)


docs/framework-introscope_agent.md
```diff
@@ -23,6 +23,7 @@ The credential payload of the service may contain the following entries:
 
 | Name | Description
 | ---- | -----------
+|`agent_defaultProcessName`| (Optional) The name that is specified for the agent process. If not specified, default value is the application name.
```
nebhale: You're comfortable that this doesn't appear to be consistent with other names in this payload (use of camel case)?

DMevada: Yes, I am fine that because we decided that adding extra underscores means more typing for the user.

nebhale: And the inconsistency more difficult to predict.  But that's fine.

PR #632, opened 2018-09-13
Updated the appmon version and download url

arthfl:
* Updated the download url, since the old one will be deprecated soon
* updated the agent version


config/dynatrace_appmon_agent.yml
```diff
@@ -15,6 +15,6 @@
 
 # Configuration for the Dynatrace framework
 ---
-version: 6.3.0_+
-repository_root: http://downloads.dynatracesaas.com/cloudfoundry/buildpack/java
-default_agent_name: 
+version: 7.1.0_+
+repository_root: http://files.dynatrace.com/downloads/appmon/cloudfoundry/buildpack/java
```
esukram: @arthfl can we please default to *`HTTPS`* schema?

arthfl:
Changed the URL to use https instead of http.

thomas-klambauer:
@nebhale Do you have any information on when this change / version goes GA?
If possible, could we push that out this week or next, because the original URL that is updated here will go offline and it would be great if we could have a solution out to users at that point in time.

nebhale:
Went out in 4.16 yesterday.  Thanks for your patience.

thomas-klambauer:
Thanks! Good to know! we'll start dismantling the old URL.

nebhale:
I'd be very careful doing that this quickly.  Just because a buildpack is available doesn't mean that your customers are upgrading to it.  At the very least, most enterprises wait for the next monthly patch release (in this case, November's) and get the buildpack upgrade there.  Additionally, there are a number of customers, against our recommendation, that are using versions 2, 5, and even 10 versions in arrears.  I would assume that removing that URI in the next six months or so will cause disruption and outages to your customers on CF.

thomas-klambauer:
Thanks for the hint. We forewarned customers also over other channels and it will take a while more to get it fully done in any case.

PR #633, opened 2018-09-18
Set maxHttpHeaderSize in tomcat server as a parameter

hmlingesh:
Hello Team, 

Recently JWT token usage trend increased by different microservices and client entities. JWT defines no upper limit in the spec (http://www.rfc-editor.org/rfc/rfc7519.txt) we do have some operational limits. As a JWT is included in an HTTP header, we have an upper limit (SO: Maximum on HTTP header values) of 8K on the majority of current servers. Due to avoid this conflict would like to add maxHttpHeaderSize is part of buildpack, so that the entity owner can decide how much max size is allowed for their clients. 

Even though buildpack having an alternate option to modify tomcat configuration via https://github.com/cloudfoundry/java-buildpack/blob/master/docs/container-tomcat.md#external-tomcat-configuration will be good this specific parameter is part of buildpack.

Considering this necessary I would like to send this pull request.

Regards
Lingesh M

### DIFF
--- a/config/tomcat.yml
+++ b/config/tomcat.yml
@@ -20,6 +20,8 @@ tomcat:
   repository_root: "{default.repository.root}/tomcat"
   context_path: 
   external_configuration_enabled: false
+  maxHttpHeaderSize_enabled: false
+  maxHttpHeaderSize: 24576
 external_configuration:
   version: 1.+
   repository_root: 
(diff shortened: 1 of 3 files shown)


nebhale:
There hasn't been a strong request this feature beyond its original audience.  Given that this is an expansion of configuration for Tomcat and down the slippery slope of the entirety of Tomcat's configuration, I'm not going to merge it in.

PR #635, opened 2018-09-28
Allow specifying the jacoco agent's output mode to choose between tcp…

konstantinblaesi:
…client or tcpserver.

Use case:

Generating coverage reports using JaCoCo for a spring boot app.
```
cf cups jacoco-service -p '{"output":"tcpserver",address":"localhost"}'
cf bind-service APP_NAME jacoco-service
cf restage APP_NAME
cf ssh -N -L localhost:6300:localhost:6300 APP_NAME &
... run tests ...
java -jar jacococli.jar dump --address localhost --destfile jacoco.exec
... run sonar scanner ...
```

Reason:

I cannot use the tcpclient mode because the organisation disabled the `cf add-network-policy` command (no container to container networking), hence I cannot setup/use a jacoco tcp server/service in Cloud Foundry that's supposed to receive the coverage data.

### DIFF
--- a/docs/framework-jacoco_agent.md
+++ b/docs/framework-jacoco_agent.md
@@ -27,6 +27,7 @@ The credential payload of the service may contain the following entries:
 | `excludes` | (Optional) A list of class names that should be excluded from execution analysis. The list entries are separated by a colon (:) and may use wildcard characters (* and ?).
 | `includes` | (Optional) A list of class names that should be included in execution analysis. The list entries are separated by a colon (:) and may use wildcard characters (* and ?).
 | `port` | (Optional) The port for the agent to connect to
+| `output` | (Optional) The mode for the agent. Possible values are either tcpclient (the default) or tcpserver. 
 
 ## Configuration
 For general information on configuring the buildpack, including how to specify configuration values through environment variables, refer to [Configuration and Extension][].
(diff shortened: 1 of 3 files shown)


nebhale:
You don't need Container-to-Container networking to get outbound connections out of a container to a service though.  Every time your application connects to a database or makes an outbound HTTP to connect to another microservice, you make the same kind of outbound connection.

nebhale:
Since there's been no response on this issue in a couple of months, I'm going to close it.  If you'd like to see it re-opened, please comment on the issue and I'll reopen it.

konstantinblaesi:
Right outbound is not a problem, but I need a tcp server that the jacoco agent is able to connect to right? I would have probably done that if our Cloud Foundry environment didn't have Container-to-Container networking disabled at the time. I needed a solution because I didn't know whether `add-network-policy` was ever going to be available.
I updated the branch, what do you think?
[CLA should be covered](https://github.com/orgs/sap-cloudfoundry/people?utf8=%E2%9C%93&query=konstantinblaesi) ?

konstantinblaesi:
@nebhale ping (not sure if you saw my comment since the PR is closed)

konstantinblaesi:
@nebhale ping

PR #653, opened 2018-11-21
Disabled CDS via Command Line Option for Dynatrace Injection

rbamberger:
Making the Agent injection future proof for upcoming Java Version Changes

### DIFF
--- a/lib/java_buildpack/framework/dynatrace_appmon_agent.rb
+++ b/lib/java_buildpack/framework/dynatrace_appmon_agent.rb
@@ -36,6 +36,7 @@ module JavaBuildpack
       # (see JavaBuildpack::Component::BaseComponent#release)
       def release
         @droplet.java_opts.add_agentpath_with_props(agent_path, name: agent_name, server: server)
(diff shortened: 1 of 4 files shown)


nebhale:
What's the desire to do this explicitly?  At the moment there is no class data sharing generally, and it's not clear that it could ever be enabled (given the container isolation).

rbamberger:
Oracle enabled Class Data Sharing aka CDS by default for Java 11 (-Xshare:off is set to -Xshare:auto). Under certain circumstances this can lead to crashes in combination with the Dynatrace Agent. The problematic behavior of the JVM has already been corrected in the upcoming Java 12 (19.3).
A backport to Java 11 is at least planned. Adding the explicit VM argument ensures the same corrected behavior for all Java 11 versions with and without the backport.

rbamberger:
The option we set (-Xshare:off) is the default for all Java Versions < 11. So it does not change the behaviour of these versions. We have tested it in our internal Lab. Therefore I think it is not necessary to add conditionals explicitly for Java 11.

PR #657, opened 2018-11-30
Support advanced configuration via AppDynamics extension buildpack 

pavankrish123:
Added support to read extended configuration given by appdbuildpack (multi buildpack)

- The PR includes a change to read the advanced configuration files from `/home/vcap/app/.appdynamics/.` folder into `/home/vcap/app/.java_buildpack/ver*/`.

- The idea here is that the files in .appdynamics are installed through appdynamics extension supply buildpack. The files themselves are picked by appdynamics buildpack by looking in the zip, jar, war (or any artifacts that are getting pushed)

- Since ProfileD support is not there in java_buildpack - this should make do.

### DIFF
--- a/lib/java_buildpack/jre/open_jdk_like_jre.rb
+++ b/lib/java_buildpack/jre/open_jdk_like_jre.rb
@@ -64,9 +64,8 @@ module JavaBuildpack
       def release
         @droplet
           .java_opts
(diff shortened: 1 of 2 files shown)


pavankrish123:
@nebhale Please close this PR. I have submitted PR #684 that solves the majority of the scenario with the approach you have suggested. 

Thanks!

PR #662, opened 2018-12-13
read credentials to download the proper profiler zip file

fwanggg:
If our customer is using online JBP and they have a older version of boshAddon, we don't want them to unexpectedly download a newer version of JBP artifacts by default, which is the default behavior of JBP unfortunately. we want the version of JBP artifacts to be lower or equal to our bosh AddOn version.
our service broker will set version number for JBP to read. 

I feel this is not the best solution to solve the version mismatch issue. Other buildpacks, such as HWC/dotnet core all support multi-buildpack, which we don't have to worry some customers use BP from some other central repository whose version and mechanism we have no control over. 

At the moment, we can only modify JBP code not tile or bosh addon or anything in our software's code. So i hope this change can mitigate the issue temporarily until we find the perfect solution.


nebhale:
The fundamental problem here is that the BOSH add-on and agent don't have a compatibility window (for example, what happens when you update your BOSH add on, and buildpack, but _haven't_ restaged every application that uses it).  I don't view this as a problem with the Java Buildpack as it will continue to exist even in a multi-buildpack world.

I'm not comfortable with a credential payload providing a version number for resolving the dependency version.  If you want to update the broker to serve a download URI from the service broker (has to be on-prem since most customers are air-gapped) like [Dynatrace](https://github.com/cloudfoundry/java-buildpack/blob/master/lib/java_buildpack/framework/dynatrace_one_agent.rb#L103) I'd be open to making that change.

fwanggg:
@nebhale From our perspective, we allow version mismatch, we can have more recent bosh-addon and older java buildpack(with older artifacts). Our COM protocol is backward compatible. Think of bosh-addon as server and the aritifacts are clients. you can upgrade your server without forcing your customers to upgrade their clients.

Whether or not customer is aware that they are upgrading/upgraded is critical here.
And in your example, customer would only upgrade bosh-addon via tile upgrading, and in this case, customer know what they are upgrading.

The bad situation is they don't want to upgrade yet and we post new artifacts. Customer simply restage an app or push an new app. New artifacts will be downloaded but bosh-addon are still old. 

thanks for the dynatrace pointer. Due to the limitation of our broker, I can only introduce a configurable baseurl and then combine the baseurl with version through credentials to form the final download url.

fwanggg:
Also, Is there any plan to support multi-buildpack as non-final step? it would be nice that the responsibility of official buildpacks v.s. third-party vendor ext. buildpack, can be clearly separated.

fwanggg:
@nebhale Hi, Ben, i've added "baseurl" so customer can specify on-prem file server base url.
I don't see how dynatrace solves version matching problem through. looks like they always use latest and not configurable or dynamic. 
I'd like not not have customers modify configs to specify versions. They've been reluctant to do so because then they will maintain too much versions of JBP.
Please take a look at the new change. If you are ok with it, let me solve the merge conflicts and update the documentation. Let me know.

nebhale:
@fwanggg We're not going to put any more effort towards improvements in this version of the Java Buildpack (or the currently released versions of the other buildpacks).  Instead, everyone is working towards the new [Cloud Native Buildpacks](https://buildpacks.io).  We're going to reach out to the companies that have current integrations shortly to start working with them on migration.

nebhale:
Good to hear that there's some skew allowed between the agent and the BOSH add on.  I think you should be able to depend on the fact that the same user is upgrading the two of them though, and therefore can be expected to understand that the BOSH add-on *must* be upgraded before the buildpack is.

By serving the agent dependency from the service broker, Dynatrace never _needs_ to worry about version matching.  The Service Broker and their BOSH add-on are both updated in lockstep (in their tile) and therefore, they can depend on the upgrade of the agent only happening the next time a stage happens after the BOSH add-on is upgraded.

The use of `baseurl`  won't be sufficient for this contribution as it stands now.  The buildpack needs to be able work in a completely air-gapped environment and that's not how things stand now (depending on `https://pcf-instrumentation-download.steelcentral.net` if the value is unspecified).  Instead of continuing down this path, I'd *strongly* encourage you to update your service broker.  It can be as simple as adding support for `GET /agent` and streaming the correct tarball back.  Then the credential payload contains a pointer to the proper host name (determined when the broker is deployed) and the buildpack downloads from there.

chestnutt:
Hi Ben
When you say "_I'd strongly encourage you to update your service broker. It can be as simple as adding support for GET /agent and streaming the correct tarball back. Then the credential payload contains a pointer to the proper host name (determined when the broker is deployed) and the buildpack downloads from there._" -- does this mean we could include our tarball with our Tile, as part of the service broker?  And that the JBP would be able to download the tarball from there? That would be sweet!

nebhale:
@chestnutt Yep, that would be the plan.  It's the most direct way to keep both the client and the BOSH add-on in sync with one another.

fwanggg:
close the PR. we will open another one with suggested changes.

chestnutt:
@nebhale To follow up: I need to modify our Service Broker to include the zip file as it's deployed so it can serve it up.  How do I get the Tile to include the zip file with the SB deployment?

chestnutt:
@nebhale OK, Serious question: adding the url in the credentials should work - but what happens during an upgrade?  I remember something earlier about PCF caching credentials -- will that ever be a problem?

nebhale:
You should likely have a single immutable URI exposed `https://riverbed-broker.cfapps.io/agent` which only serves the "current" agent at any given time.  This does not solve the fact that the agent is baked into the droplet (which requires a restage to update), but ensures that credentials don't need to change during upgrades.  Also, you should probably be update credentials for bindings (which is totally a thing) as part of credential rotation.

chestnutt:
@nebhale Earlier, you said "_Then the credential payload contains a pointer to the proper host name (determined when the broker is deployed) and the buildpack downloads from there_"

I thought _determined when the broker is deployed_ meant to use the host name from the request header of the "/v2/service_instances" request. Is that the right way to get the host name? Or is "https://riverbed-groker.cfapps.io/agent" a magic URL that we should use instead?

nebhale:
It's not a magic host name, nor do you need to get it from the `HOST` header.  When you deploy your tile, it deploys your broker.  This deployment is likely via `cf push` at which point you choose the hostname or possibly via BOSH deployment which returns an assigned IP/hostname as well.  Either one of these is the correct configuration to use.

nebhale:
Note that inside of a CF container (`cf push` style, not BOSH style) your routes are exposed in an environment variable that you can query on startup.

image

chestnutt:
I use tile-generator, so I don't choose a hostname myself.  How do I find out what it's using?

chestnutt:
@nebhale So you're saying the HOST header is wrong and there's an environment variable I should get the name from?

nebhale:
It can be right, but it's not guaranteed to be.  For example the Cloud Controller that is asking the broker to generate credentials may address the broker via an IP address that is not routable from applications.  The truth is always what the platform tells the application.

chestnutt:
@nebhale  So, the ServiceBroker should pull the URI from the `application_uris` in `VCAP_APPLICATION`? Is that the correct one to use??

nebhale:
I _think_ that's the right one.  It could also be `uris` but generally we only go towards more-specific and hold the original for backwards compatibility.  Checking in with the partner team might tell you which is the "preferred" variant of those two.

PR #668, opened 2019-01-03
Provide example for `headroom` option

oppegard:
This replaces PR #667 since I've already signed the CF CLA.


nebhale:
Resolved by db7771a.

PR #681, opened 2019-01-31
download artifact from service broker

fwanggg:
@nebhale  per your earlier suggestion, we've made our service broker providing a URL to stream the artifact down to the JBP. Our service broker will always supply a URL through credentials for JBP to consume. JBP will now by default download artifacts from that new URL in credentials. If anything bad happens, it will fall back and use the JBP's official default approach.


chestnutt:
@nebhale Hi Ben - any chance of getting this merged for the next release of JBP?  We're ready to use it in our next beta.
Thanks -Dave

fwanggg:
@nebhale great talk today. i've added some notes to our doc regarding to mechanism change starting from our 10.20 service broker per our discussion.

PR #684, opened 2019-02-07
Provide default configuration to mask VCAP_* env variables in AppDynamics Controller 

pavankrish123:
- This change starts the appdynamics java agent with default configuration (checked into `resources/app_dynamics_agent/defaults/conf` directory.

- `app-agent-config.xml` can only be present in `ver*` directories of the appdynamics agent. So it is now coped to all ver* folders present.

- customers can still over-ride this configuration using existing mechanism

- @nebhale please close PR #657 . This default configuration covers what many (all) the customers who would like to do through that mechanism. 

- We will tackle advanced custom configuration in a separate PR

### DIFF
--- a/resources/app_dynamics_agent/defaults/conf/app-agent-config.xml
+++ b/resources/app_dynamics_agent/defaults/conf/app-agent-config.xml
@@ -1,0 +1,702 @@
+<app-agent-configuration>
+    <configuration-properties>
+        <property name="agent-overwrite" value="false"/>
+        <!-- enables/disables the agent. TransactionEntryPoints will not be monitored. No new bt and metrics will be registered;
+        metrics, snpshots will not be reported; The background threads will not be stopped and once this is turned back on
+        monitoring is active immediately. Does not need a restart -->
+        <!-- property name="disable-agent" value="true"/ -->
+
+        <!-- defines the interval in seconds for the agent to poll for new configuration from the controller. needs a restart -->
+        <!-- property name="config-poll-interval" value="60"/ -->
+
+        <!-- uncomment this to enable timing stats for agent and controller rest communication. does not need a restart -->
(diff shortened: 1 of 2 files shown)


resources/app_dynamics_agent/defaults/conf/app-agent-config.xml
```diff
@@ -0,0 +1,702 @@
+<app-agent-configuration>
```
nebhale: This file isn't quite the equivalent of the contents of the download with one stanza added.  I'll make a polishing commit to bring it inline.

pavankrish123: Thanks @nebhale : Please make sure to get PR#699 as well.

The filter should be 

```
        <sensitive-data-filter applies-to="environment-variables,system-properties"
                               match-type="STARTSWITH"
                               match-pattern=""/>
```
not 

```
        <sensitive-data-filter applies-to="environment-variables,system-properties"
                               match-type="STARTSWITH"
                               match-pattern="*"/>
```

resources/app_dynamics_agent/defaults/conf/app-agent-config.xml
```diff
@@ -0,0 +1,702 @@
...
+                               match-type="CONTAINS"
+                               match-pattern="password"/>
+
+        <sensitive-data-filter applies-to="environment-variables,system-properties"
+                               match-type="CONTAINS"
+                               match-pattern="key"/>
+        
+        <sensitive-data-filter applies-to="environment-variables"
+                                    match-type="STARTSWITH"
+                                    match-pattern="*"/>
```
pavankrish123: this should be 
```
        <sensitive-data-filter applies-to="environment-variables,system-properties"
                               match-type="STARTSWITH"
                               match-pattern=""/>
```

PR #699, opened 2019-03-12
Use empty string "" filter  not * to mask all env variables for AppD 

pavankrish123:


nebhale:
This masking of all environment variables is quite different than all other APM providers and I'm not sure it's what customers expect.

pavankrish123:
@nebhale - I share the same thoughts on masking everything. However from the ongoing discussion, the team felt it's better to secure everything by default and then open up per requirement.  In any case, I will put in another PR to dynamically fetch the `app-agent.xml` to over ride the default behavior as per your original suggestion

PR #706, opened 2019-04-01
Add support for exporting any java agent property from the credential…

DMevada:
…s. Before this commit, exporting a new property meant adding a new method to read the property key from the credentials and then setting it as a java option. After this commit, all properties that are in the credentials will be set as a java option, without the need for adding code to read each individual property. The reason for this change is to keep the introscope framework concise, and it does not need to be updated in the future for new properties.


DMevada:
Please ignore this pull request, I need to make some more changes. I will submit a new one soon.

Issue #712, opened 2019-04-04
Frameworks Cause Application Crashes when Space Name Includes Whitespace


owenfarrell:
The AppDynamics, Contrast Security, and SkyWalking agents are 3 frameworks that inject the space name in to JAVA_OPTS.

In the event that a space name includes whitespace (`My Dev Space`), the result is unintentional whitespace getting added to the command line. The additional whitespace means Java is unable to identify the appropriate argument for the main class.

A quick fix here would be to update the default configuration attributes for the above frameworks, ensuring that the end result is escaped in shell-safe manner.

A more robust fix would entail sanitizing the space name any time it is used in an O/S level operation (e.g. JAVA_OPTS, file paths?)

PR #713, opened 2019-04-04
Frameworks Cause Application Crashes when Space Name Includes Whitespace

owenfarrell:
Proposal for #712

### DIFF
--- a/config/app_dynamics_agent.yml
+++ b/config/app_dynamics_agent.yml
@@ -17,6 +17,6 @@
 ---
 version: 4.+
 repository_root: https://packages.appdynamics.com/java
-default_application_name: $(jq -r -n "$VCAP_APPLICATION | .space_name + \":\" + .application_name")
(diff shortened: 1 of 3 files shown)


owenfarrell:
@nebhale - So i think my first crack at this could serve as a quick fix for the existing frameworks that put the space name on the command line. But if the `@configuration` is blanked out via a buildpack fork or `JBP_CONFIG_` environmental variable, then the fallback value generated by these same frameworks will cause a repeat of the same issue.

Updating the `qualify_value` method in `java_buildpack/component/java_opts` [(here)](https://github.com/cloudfoundry/java-buildpack/blob/master/lib/java_buildpack/component/java_opts.rb#L125) might be a more universal fix that covers command line arguments generated by either a configuration or a framework. Something like this could prevent extra whitespace, while preserving the injection of environmental variables.

```ruby
def qualify_value(value)
  value.respond_to?(:relative_path_from) ? qualify_path(value) : "\"#{value}\""
end
```

But making that change has pretty significant implications on the unit tests.


Thoughts?

PR #737, opened 2019-07-03
Add support to parse newly added access key secret field 

pavankrish123:
-  Starting 4.7 service broker tile, we are introducing secret typed form  fields for sensitive data  (account access key) to replace text typed fields. This will lead to a different VCAP_SERVICES structure and hence requires additional parsing logic. It is a one line change to honor the secret typed field

- Please note that we are going to deprecate the older text field eventually (2 months) but till then we need to support the text field as well.

this is what new credentials structure look like
```
          "credentials": {
            "account-access-key": null,
            "account-access-secret": {
              "secret": "60dbc1a8-043c-44f7-93f8-e18aeb0b8274"
            },```

- Have also added UNIQUE_HOST_ID in the config for our container correlation


pavankrish123:
Stale PR -Closing it

PR #741, opened 2019-08-15
Seeker Agent integration with pivotal

hbarshak:
Hi Pivotal,

This patch provides the required configuration for integrating java based web application with the Seeker agent in order to detect security vulnerabilities.
This PR relates to the Seeker tile: https://network.pivotal.io/products/synopsys-seeker

More info about Seeker can be found here: https://www.synopsys.com/software-integrity/security-testing/interactive-application-security-testing.html


nebhale:
@hbarshak Thanks for the contribution.  Do you have any idea how far back the ability to download your agent from the service broker goes?  If it predates (as in customer would have had access to it) before I get this merged in, then I'd like to simply the contribution so that it assumes you'll download from the service broker.

Does that seem viable to you?

PR #744, opened 2019-09-06
Revert "Added default_server_name to Contrast Security Framework"

gmoore:
This reverts commit cd849e8ba7d385946f4e99233c2525188fcb6e20 from this Pull Request https://github.com/cloudfoundry/java-buildpack/pull/702

cd849e8 introduced a default server naming convention that was in conflict with the ways that others were setting server names. We would like to return to the server naming behavior as it was before cd849e8. We've discussed this change with @owenfarrell

### DIFF
--- a/config/contrast_security_agent.yml
+++ b/config/contrast_security_agent.yml
@@ -16,5 +16,4 @@
 # Configuration for the ContrastSecurity framework
 ---
 version: 3.+
-default_server_name: $(jq -r -n "$VCAP_APPLICATION | .space_name + \":\" + .application_name | @sh"):$CF_INSTANCE_INDEX
 repository_root: https://artifacts.contrastsecurity.com/agents/java
--- a/docs/framework-contrast_security_agent.md
+++ b/docs/framework-contrast_security_agent.md
@@ -27,7 +27,6 @@ The framework can be configured by modifying the [`config/contrast_security_agen
 | Name | Description
 | ---- | -----------
 | `repository_root` | The URL of the Contrast Security repository index ([details][repositories]).
-| `default_server_name` | The default server name for this application in the Contrast dashboard. The default value is an expression that will be evaluated based on the `space_name`, `application_name`, and `instance_index` of the application. |
 | `version` | The version of Contrast Security to use. Candidate versions can be found in [this listing][].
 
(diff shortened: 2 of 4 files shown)


owenfarrell:
@gmoore - As discussed over email, I struggle with the concept of taking **_away_** the de facto standard that CloudFoundry has established.

> cd849e8 introduced a default server naming convention that was in conflict with the ways that others were setting server names

Is that because the Contrast instrumentation agent considers Java system properties to have the highest order of precedent? If so, I think there are 2 alternatives that could maintain the standard, but also address your concern:

1.  set the server name to a different location (i.e. environmental variable) with a lower order of precedence
1. do a bettter job of checking for the properties that are getting clobberd. If I'm reading the code right, parsing JAVA_OPTS comes  **_after_** this framework is run ([component list](https://github.com/cloudfoundry/java-buildpack/blob/master/config/components.yml)). So checking the contents of JAVA_OPTS (in `appname_exist?` and `server_exist?`) will always come up empty.


I would need some guidance to make sure I'm reading this right, but my local experiments seem to confirm that behaviro.

gilday:
Hi @owenfarrell, I am one of the Contrast Java agent developers, and I can help us find a solution that works for everyone.

The problem with the code that is being reverted here is that it changes the default server name _unless the user has provided a server name override_. The issue with this approach is that [Contrast offers users many ways of overriding the server name](https://docs.contrastsecurity.com/installation-javaconfig.html); so, the buildpack would need to know about all of these ways and basically repeat the config lookup code that is built into the agent in order to check whether or not the user has explicitly provided a server name.

As written, the buildpack checks to see if the system property `contrast.server` is set. The property `contrast.server` is a legacy property name that Contrast still supports for backwards compatibility. The agent also allows users to specify the server name using

* system property `contrast.server.name`
* external system properties file
* env var `CONTRAST__SERVER__NAME`
* YAML config file (which may be at some default location or alternatively specified by env var `CONTRAST_CONFIG_PATH` or system property `contrast.config.path`)

It would be prohibitively hard to keep the buildpack logic in sync with all this config lookup logic to determine if the user is explicitly specifying a server name. Instead, could we add a configuration option to the buildpack itself that would allow PCF users to opt-in or out of the buildpack's `default_server_name` logic and specify their own server name?

PR #751, opened 2019-10-17
Enable configuration override during compile phase if a config server is specified via manifest

hp685:
Enables agent log messages to be routed to stdout by overriding agent and log config during compile phase


hp685:
I signed the CLA

hp685:
I'll create a PR from Appdynamics repo since that has CLA auth working

PR #769, opened 2020-01-19
Fix line break for Oracle cf set-env example and unify bash markdown …

mayrstefan:
…for all JREs


docs/jre-zulu_jre.md
```diff
@@ -25,9 +25,10 @@ The JRE can be configured by modifying the [`config/zulu_jre.yml`][] file in the
 
 To use Zulu JRE instead of OpenJDK without forking java-buildpack, set environment variable:
```
nebhale: Missed one final ` and restage`, but I'll get it when I merge 😄.

Issue #770, opened 2020-01-21
ConfigurationUtils Test Failure


vipinvkmenon:
on running `bundle exec rake` the tests fail with the error.

```
Failures:

  1) JavaBuildpack::Util::ConfigurationUtils   diagnoses invalid YAML syntax
     Failure/Error:
       expect { described_class.load('test') }
         .to raise_error(/User configuration value in environment variable JBP_CONFIG_TEST has invalid syntax/)

       expected /User configuration value in environment variable JBP_CONFIG_TEST has invalid syntax/ but nothing was raised
     # ./spec/java_buildpack/util/configuration_utils_spec.rb:122:in `block (4 levels) in <top (required)>'

Finished in 38.17 seconds (files took 1.37 seconds to load)
895 examples, 1 failure

Failed examples:

rspec ./spec/java_buildpack/util/configuration_utils_spec.rb:121 # JavaBuildpack::Util::ConfigurationUtils   diagnoses invalid YAML syntax

Randomized with seed 19818

/usr/local/Cellar/ruby/2.6.4_2/bin/ruby -I/usr/local/lib/ruby/gems/2.6.0/gems/rspec-core-3.9.0/lib:/usr/local/lib/ruby/gems/2.6.0/gems/rspec-support-3.9.0/lib /usr/local/lib/ruby/gems/2.6.0/gems/rspec-core-3.9.0/exe/rspec --pattern spec/\*\*\{,/\*/\*\*\}/\*_spec.rb failed
```

nebhale:
I'm unable to reproduce this failure, either [in CI](https://java-experience.ci.springapps.io/teams/java-experience/pipelines/java-buildpack/jobs/java-unit-2.6/builds/87) or locally:

```
➜  ~/d/s/java-buildpack master rbenv shell 2.6.5
➜  ~/d/s/java-buildpack master bundle install
Using rake 13.0.0
Using public_suffix 4.0.1
Using addressable 2.7.0
Using ast 2.4.0
Using bundler 2.0.2
Using safe_yaml 1.0.5
Using crack 0.4.3
Using diff-lcs 1.3
Using hashdiff 1.0.0
Using jaro_winkler 1.5.3
Using parallel 1.18.0
Using parser 2.6.5.0
Using rainbow 3.0.0
Using redcarpet 3.5.0
Using rspec-support 3.9.0
Using rspec-core 3.9.0
Using rspec-expectations 3.9.0
Using rspec-mocks 3.9.0
[...]

PR #771, opened 2020-01-21
Buildpack test fails. Issue #770

vipinvkmenon:
The test to diagnose invalid YAML fails. This commit fixes the test case by correcting JBP_CONFIG_TEST.

Resolves: #770


nebhale:
Based on the discussion in #770 I don't think that we'll be making any changes to the test.  I am still interested in determining why you aren't seeing the same behavior.

PR #776, opened 2020-01-29
Addition of Jaeger-Agent

vipinvkmenon:
This change adds support for Jaeger-Agent.


nebhale:
Thanks for the contribution.  I've been churning over it for a while and unfortunately, I don't think that the timing is right for inclusion.  There are two fundamental issues with running a second process within the same container on Cloud Foundry (as it exists today).

1. The Memory Calculator that is contributed by the buildpack is designed to occupy every possible byte available within the container.  Running a second process within the container occupies some memory and at max load, while likely tip the container over its limit causing Diego's enforcement mechanism to recycle the container at the worst possible time.  This _is_ solvable (you can convince the memory calculator to carve out some space for a second process), but we've never done it directly before and if this was the only problem, we'd happily take the contribution and work it out.
1. There is only a single health check within the container.  So by starting the Jaeger process in the background, Diego is not watching it to keep it alive.  Now, I'm guessing that the application will continue to work as expected even if it can't access the Jaeger agent, but the contract that Cloud Foundry has with its users is that a container is fully healthy or will be restarted.  If a user expects to get metrics out of Jaeger, but can't then we've broken our contract.  Heck, even if they knew which instance had the healthy application but an unhealthy Jaeger, we give the user no way to restarting that specific instance without restarting every instance and possibly incurring down time.

Both of those issues have been identified in the past and Cloud Foundry is adding support for side-car processes within an application's container.  This facility addresses both issues (independent memory limits, independent healthchecks) and would be a much better way to deliver this functionality to users.

PR #782, opened 2020-03-02
Switch back to normal downloading of our profiler

chestnutt:
No longer downloaded from the Service Broker. This is necessary for customers to be able to create offline buildpacks for various versions.  They want to do that so that they can upgrade applications one by one to the new version of our software, after their internal testing.


chestnutt:
CLA is signed.

chestnutt:
This is a duplicate. Closing.

Issue #828, opened 2020-10-15
Unable to deploy to Cloud Foundry using Java buildpack for Dynatrace OneAgent because start_command exceeds maximum character length


gidad:
`DT_CONNECTION_POINT` parameter contains a list of endpoints and sometimes can be between around 3000 and 4000 characters long.

That variable should be exported (e.g. with `cf set-env`)  before calling `start_command` so that we can shorten the `start_command` string.

Here is a `start_command` sample which generates the issue:

```
start_command:
JAVA_OPTS="-agentpath:$PWD/.java-buildpack/open_jdk_jre/bin/jvmkill-1.16.0_RELEASE=printHeapHistogram=1 -Djava.io.tmpdir=$TMPDIR -XX:ActiveProcessorCount=$(nproc)
-agentpath:$PWD/.java-buildpack/dynatrace_one_agent/agent/bin/linux-x86-64/liboneagentloader.so -Xshare:off
-Djava.ext.dirs=$PWD/.java-buildpack/container_security_provider:$PWD/.java-buildpack/open_jdk_jre/lib/ext -Djava.security.properties=$PWD/.java-buildpack/java_security/java.security $JAVA_OPTS" &&
CALCULATED_MEMORY=$($PWD/.java-buildpack/open_jdk_jre/bin/java-buildpack-memory-calculator-3.13.0_RELEASE -totMemory=$MEMORY_LIMIT -loadedClasses=12824 -poolType=metaspace -stackThreads=250
-vmOptions="$JAVA_OPTS") && echo JVM Memory Configuration: $CALCULATED_MEMORY && JAVA_OPTS="$JAVA_OPTS $CALCULATED_MEMORY" && MALLOC_ARENA_MAX=2 DT_TENANT=xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxx
DT_TENANTTOKEN=xxxxxxxxxxxxxxxxxx
DT_CONNECTION_POINT="<...3650_characters_of_urls...>"
DT_APPLICATIONID=test-api DT_LOGSTREAM=stdout SERVER_PORT=$PORT eval exec $PWD/.java-buildpack/open_jdk_jre/bin/java $JAVA_OPTS -cp $PWD/. org.springframework.boot.loader.JarLauncher
```

TimGerlach:
@aloismayr fyi

aloismayr:
@mmayr-at @mreider fyi

mmayr-at:
[...]

PR #830, opened 2020-10-23
Add configuration option to overwrite communication endpoints for Dynatrace OneAgent

mmayr-at:
This is another solution to #828 where a user can specify the communication endpoints for the Dynatrace OneAgent via an env variable or in the application manifest.yaml, e..g:
```yaml
---
applications:
- name: <application>
   ...
services:
- dynatrace-service
env:
    JBP_CONFIG_DYNATRACE_ONE_AGENT: '{ communication_endpoints: "https://test1;https://test1" }'
```
or use 
```
cf set-env <application> JBP_CONFIG_DYNATRACE_ONE_AGENT '{ communication_endpoints: "https://test1;https://test1" }'
```

### DIFF
--- a/config/dynatrace_one_agent.yml
+++ b/config/dynatrace_one_agent.yml
@@ -1,0 +1,18 @@
+# Cloud Foundry Java Buildpack
+# Copyright 2013-2020 the original author or authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Configuration for the Dynatrace OneAgent framework
+---
+communication_endpoints:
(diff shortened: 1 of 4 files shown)


TimGerlach:
Hi @mmayr-at, I am not sure if this is a very user-friendly fix. This requires the developer to get an up-to-date list of communication endpoints before deploying their apps to Cloud Foundry. I don't think that an end user should even deal with some IP lists that they got from anywhere.

nebhale:
@mmayr-at It's not clear to me that this solves the problem that a command line can be too long, other than to artifically reduce the list of connection points by hand.  Is there a better solution to limit the list coming from the server, or to encourage the buildpack to only pick the first N number of elements?

mmayr-at:
Hi @TimGerlach, I totally agree with you, that this is not really user-friendly. However this fix can be considered as last resort to make a deployment working, because there is currently no other way to limit or shorten the list of communication endpoints on the server side.

When the buildpack picks only the first N number of entries, this can introduce another problem. Let's assume only the last communication endpoint is reachable, then using the first N entries would remove the only working endpoint and the agent can't connect to the Dynatrace environment. A solution for this are [network zones](https://www.dynatrace.com/support/help/setup-and-configuration/network-zones/), unfortunately they aren't available in the deployment API yet.

tehall16:
@dmikusa-pivotal this might need to be reviewed again, came up for a customer escalation

dmikusa-pivotal:
Hi all, is this still something that we are actively pursuing? 

I see #893 (and #894) which may be doing something similar? I'm not certain as I don't know Dynatrace that well.

If this is something that anyone involved is still interested in working on please let me know by the end of July 2021, otherwise I will go ahead and close out this PR. Thanks!

mmayr-at:
Hi @dmikusa-pivotal, after #893 and #894 this PR is not necessary anymore and we can close it. With the full support of the network zone in the buildpack and Dynatrace API a user can control the length of the endpoint list directly in the Dynatrace product and does not need to use this solution.

PR #840, opened 2020-11-25
add option to override AppDynamics config from artifact

vchrisb:
Added `APPD_CONF_DIR` environment variable support to copy configuration files from this directory, relative to the application root dir, to override the default AppDynamics config e.g. `BOOT-INF/classes/appdynamics/conf`.
This is implemented in the same way as `APPD_CONF_HTTP_URL`

If both `APPD_CONF_HTTP_URL` and `APPD_CONF_DIR` are configured, both source will be used but config files from `APPD_CONF_DIR` will have preference.


spec/java_buildpack/framework/app_dynamics_agent_spec.rb
```diff
@@ -165,6 +165,18 @@
           component.compile
         end
       end
+
+      context do
+
+        let(:environment) { { 'APPD_CONF_DIR' => 'BOOT-INF/classes/appdynamics/conf' } }
```
ekcasey, line 171: I am not sure this test is really testing anything. For example, if I change this path to `BOOT-INF/classes/appdynamics/foo`, the test still passes. This is happening because the `stub-app-dynamics-agent.zip` fixture does not have a version directory.

To ensure this functionality works I think we need to add a directory like `ver1.2.3` to the text fixture and then assert that the provided `app-agent-config.xml` is copied into the version directory.

For example, if we changed the contents of `spec/fixtures/framework_app_dynamics_agent/BOOT-INF/classes/appdynamics/conf/app-agent-config.xml` to a simple test string we could add an assertion along the lines of
```
expect(IO.read(sandbox + "ver1.2.3"+"conf"+"app-agent-config.xml")).to eq("app-agent-config-from-dir")
```

lib/java_buildpack/framework/app_dynamics_agent.rb
```diff
@@ -193,6 +194,24 @@ def override_default_config_if_applicable
           end
         end
       end
+
+      # Check for configuration files locally. If found, copy to conf dir under each ver* dir
+      # @return [Void]
+      def override_default_config_local
+        return unless @application.environment['APPD_CONF_DIR']
+
+        app_conf_dir = @application.environment['APPD_CONF_DIR']
```
ekcasey, line 203: I think we should check that `app_conf_dir` exists. If a user provided this environment variable I would assume that the directory should exist and a failure might be more helpful than a silently ignoring the config if, for example, they typoed the dir name.

ekcasey (approved):

PR #842, opened 2020-12-15
APMDEV-5014: Update the documentation for the Aternity java buildpack

scottmmendel:
I have added the new environment variables that have been introduced in 11.X. -Scott

### DIFF
--- a/docs/framework-riverbed_appinternals_agent.md
+++ b/docs/framework-riverbed_appinternals_agent.md
@@ -20,16 +20,19 @@ When binding Appinternals using a user-provided service, it must have <code>appi
 | ---- | -----------
 | `rvbd_dsa_port` | (Optional)The AppInternals agent (DSA) port (default 2111).
 | `rvbd_agent_port` | (Optional) The AppInternals agent socket port (default 7073).
-| `rvbd_moniker` | (Optional) A custom name for the application (default supplied by agent process discovery).
+| `rvbd_analysis_server` | (Optional) The AppInternals secure socket IP address for the Analysis server.
+| `rvbd_analysis_server_port` | (Optional) The AppInternals agent secure socket port (default 7090).
+| `rvbd_app_instance` | (Optional) A custom name for the application (default supplied by agent process discovery).
+| `rvbd_app_config` | (Optional) A custom name for the application config file (default "deafult+config.json").
 
 **NOTE**
 
-Change `rvbd_dsa_port` and `rvbd_agent_port` only if there is a port conflict
+Change `rvbd_dsa_port`, `rvbd_agent_port`, and `rvbd_analysis_server_port` only if there is a port conflict
 
 ### Example: Creating Riverbed Appinternals User-Provided Service Payload
 
 ``` 
-cf cups spring-music-appinternals -p '{"rvbd_dsa_port":"9999","rvbd_moniker":"my_app"}'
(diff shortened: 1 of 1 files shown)


docs/framework-riverbed_appinternals_agent.md
```diff
@@ -20,16 +20,19 @@ When binding Appinternals using a user-provided service, it must have <code>appi
 | ---- | -----------
 | `rvbd_dsa_port` | (Optional)The AppInternals agent (DSA) port (default 2111).
 | `rvbd_agent_port` | (Optional) The AppInternals agent socket port (default 7073).
-| `rvbd_moniker` | (Optional) A custom name for the application (default supplied by agent process discovery).
+| `rvbd_analysis_server` | (Optional) The AppInternals secure socket IP address for the Analysis server.
+| `rvbd_analysis_server_port` | (Optional) The AppInternals agent secure socket port (default 7090).
+| `rvbd_app_instance` | (Optional) A custom name for the application (default supplied by agent process discovery).
+| `rvbd_app_config` | (Optional) A custom name for the application config file (default "deafult+config.json").
```
ekcasey, line 26: It don't see these value being read by the buildpack.

ekcasey:
@scottmmendel Can you link me to the buildpack change that this PR is documenting? I don't see any recent changes in the environment variables set by the buildpack https://github.com/cloudfoundry/java-buildpack/blob/f888b34fbe93671ea74c003bea91cda8bd4bb21a/lib/java_buildpack/framework/riverbed_appinternals_agent.rb#L37-L41

scottmmendel:
These are the environment variables that can be set that our profiler consumes. Some of these are set by the buildpack itself, others can be added in the manifest.yaml file. If this file was to only document on the variables set by the buildpack, then you can delete this pull request. -Scott

ekcasey:
@scottmmendel the documentation this PR is changing describes the credentials that the buildpack receives from a [bound service instance](https://docs.pivotal.io/partners/riverbed-appinternals/using.html#bind) or [user-provided service](https://docs.cloudfoundry.org/devguide/services/user-provided.html).

The buildpack reads values from the documented credentials JSON and set equivalent environment variables. E.g if the credentials payload is `'{"rvbd_agent_port":"9999"}'` the buildpack will set `RVBD_AGENT_PORT` in the application environment. If we want users to be able to configure additional environment variables like `RVBD_ANALYSIS_SERVER` by binding a service instance we will need to update the buildpack.

Are new values like `rvbd_analysis_server` being provided in the service binding by the service broker?

Additionally, users can set any environment variable they wish in `manifest.yml`. Generally, configuration values that provide the location and credentials of an external service should be provided in the service binding. Configuration values that are not-secret and specific to the app should be provided in `manifest.yml`. I am guess that some of these new properties fall into the first bucket and some into the second. E.g Users probably want the binding to provide the correct value of `RVBD _ANALYSIS_SERVER` but `RVBD_APP_INSTANCE` is specific to the app and should be set in `manifest.yml`.

Are my assumptions correct here? If you let me know which if any of these new values are provided (or should be provided) in the service binding and I can update the buildpack.

For environment variables that should be set in `manifest.yml` you probably want to update [these docs](https://docs.pivotal.io/partners/riverbed-appinternals/using.html).

scottmmendel:
The new variables are not provided by the service broker. I will cancel the pull request and figure out how to update ".../riverbed-appinternals/using.html". -Scott

PR #843, opened 2020-12-15
Update geode-store version

sabbey37:
- Update version of Geode dependencies being pulled in by the buildpack
- Update version of Tomcat module used by Geode store
- Removed unnecessary client functions added by Geode store

### DIFF
--- a/config/tomcat.yml
+++ b/config/tomcat.yml
@@ -41,4 +41,4 @@ redis_store:
   connection_pool_size: 2
 geode_store:
   version: 1.13.+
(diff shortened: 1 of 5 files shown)


pivotal-jbarrett (approved):

ekcasey:
@sabbey37 Looks good! Have you validated that this works with real CF and Tanzu GemFire for VMs instances?

sabbey37:
> @sabbey37 Looks good! Have you validated that this works with real CF and Tanzu GemFire for VMs instances?

@ekcasey Thanks for reviewing this!  We have verified it works properly with the specified version of TGF4VMs.

ekcasey (approved):

PR #849, opened 2021-01-27
Update geode store version

sabbey37:
We would like to update the Geode store version that the Java buildpack pulls in to an earlier version so that it is compatible with more versions of TGF4VMs.

### DIFF
--- a/config/tomcat.yml
+++ b/config/tomcat.yml
@@ -40,5 +40,5 @@ redis_store:
   timeout: 2000
   connection_pool_size: 2
 geode_store:
-  version: 1.13.+
+  version: 1.11.+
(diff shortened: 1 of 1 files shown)


BenjaminPerryRoss (approved):

ekcasey:
@sabbey37 what is the urgency level for shipping this change? Also I am assuming all the other change in https://github.com/cloudfoundry/java-buildpack/pull/843 work with this older version?

ekcasey (approved):

sabbey37:
@ekcasey  Shipping the change is not urgent.  Yes, all these changes work with the changes from #843 Thank you for checking!

PR #850, opened 2021-01-27
Add basic auth support when using APPD_CONF_HTTP_URL to fetch configuration from a remote server.

dmikusa-pivotal:
Currently, if you attempt to fetch App Dynamics configuration from a remote server using `APPD_CONF_HTTP_URL`, it will work so long as the server is not protected with basic auth.

Downloads from the remote server are OK, but this [call to check_if_resource_exists](https://github.com/cloudfoundry/java-buildpack/blob/main/lib/java_buildpack/framework/app_dynamics_agent.rb#L146) doesn't use the standard download code and makes a direct HEAD request which doesn't currently support basic auth.

This patch uses basic auth for the HEAD request, which should allow it to support App Dynamics config from a remote server protected with basic auth.

### DIFF
--- a/lib/java_buildpack/framework/app_dynamics_agent.rb
+++ b/lib/java_buildpack/framework/app_dynamics_agent.rb
@@ -148,7 +148,11 @@ module JavaBuildpack
         begin
           opts = { use_ssl: true } if resource_uri.scheme == 'https'
           response = Net::HTTP.start(resource_uri.host, resource_uri.port, opts) do |http|
-            http.request_head(resource_uri)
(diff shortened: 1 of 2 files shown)


ekcasey (approved):

ekcasey:
thanks @dmikusa-pivotal !

ekcasey:
@dmikusa-pivotal would you mind signing the CLA?

dmikusa-pivotal:
Sorry, I had yet to go through the new CLA process. All done now. Thanks!

PR #851, opened 2021-01-29
Remove security credentials for geode store from JAVA_OPTS

sabbey37:
`JAVA_OPTS` prints security credentials for gemfire in the output for `cf push`.  To avoid that, we are obtaining the credentials directly from VCAP_SERVICES in the geode client auth module.

### DIFF
--- a/lib/java_buildpack/container/tomcat/tomcat_geode_store.rb
+++ b/lib/java_buildpack/container/tomcat/tomcat_geode_store.rb
@@ -71,10 +71,6 @@ module JavaBuildpack
                        :CACHE_CLIENT_LISTENER_CLASS_NAME, :SCHEMA_URL, :SCHEMA_INSTANCE_URL, :SCHEMA_LOCATION,
                        :LOCATOR_REGEXP
 
-      def cluster_operator?(user)
-        user['username'] == 'cluster_operator' || user['roles'] && (user['roles'].include? 'cluster_operator')
-      end
-
       def add_client_cache(document)
         client_cache = document.add_element 'client-cache',
                                             'xmlns' => SCHEMA_URL,
--- a/lib/java_buildpack/container/tomcat/tomcat_geode_store.rb
+++ b/lib/java_buildpack/container/tomcat/tomcat_geode_store.rb
@@ -44,6 +44,9 @@ module JavaBuildpack
         credentials = @application.services.find_service(FILTER, KEY_LOCATORS, KEY_USERS)['credentials']
         user = credentials[KEY_USERS].find { |u| cluster_operator?(u) }
 
+        File.open(@droplet.root + "gemfire.properties", "w") {|f| f.write("security-username=" + user['username'] +
+                                                                            "\nsecurity-password=" + user['password']) }
+
         @droplet.java_opts.add_system_property 'gemfire.security-username', user['username']
         @droplet.java_opts.add_system_property 'gemfire.security-password', user['password']
         @droplet.java_opts.add_system_property 'gemfire.security-client-auth-init',
--- a/lib/java_buildpack/container/tomcat/tomcat_geode_store.rb
+++ b/lib/java_buildpack/container/tomcat/tomcat_geode_store.rb
@@ -44,8 +44,9 @@ module JavaBuildpack
         credentials = @application.services.find_service(FILTER, KEY_LOCATORS, KEY_USERS)['credentials']
         user = credentials[KEY_USERS].find { |u| cluster_operator?(u) }
 
-        File.open(@droplet.root + "gemfire.properties", "w") {|f| f.write("security-username=" + user['username'] +
-                                                                            "\nsecurity-password=" + user['password']) }
+        File.open(@droplet.root + 'gemfire.properties', 'w') do |f|
+          f.write('security-username=' + user['username'] + "\nsecurity-password=" + user['password'])
+        end
 
         @droplet.java_opts.add_system_property 'gemfire.security-client-auth-init',
                                                'io.pivotal.cloudcache.ClientAuthInitialize.create'
--- a/lib/java_buildpack/container/tomcat/tomcat_geode_store.rb
+++ b/lib/java_buildpack/container/tomcat/tomcat_geode_store.rb
@@ -41,13 +41,6 @@ module JavaBuildpack
       def release
         return unless supports?
 
-        credentials = @application.services.find_service(FILTER, KEY_LOCATORS, KEY_USERS)['credentials']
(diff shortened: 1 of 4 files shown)


BenjaminPerryRoss (approved):

lib/java_buildpack/container/tomcat/tomcat_geode_store.rb
```diff
@@ -44,8 +44,10 @@ def release
         credentials = @application.services.find_service(FILTER, KEY_LOCATORS, KEY_USERS)['credentials']
         user = credentials[KEY_USERS].find { |u| cluster_operator?(u) }
 
-        @droplet.java_opts.add_system_property 'gemfire.security-username', user['username']
-        @droplet.java_opts.add_system_property 'gemfire.security-password', user['password']
+        File.open(@droplet.root + 'gemfire.properties', 'w') do |f|
+          f.write('security-username=' + user['username'] + "\nsecurity-password=" + user['password'])
+        end
```
ekcasey, line 49: Now that we are modify the filesystem I think we should move this logic from `release` to `compile`. The organization is described in these comment on `JavaBuildpack::Component::BaseComponent`

https://github.com/cloudfoundry/java-buildpack/blob/main/lib/java_buildpack/component/base_component.rb#L58-L79

While we are at it we should probably move this into a a private methods called `create_gemfire_properties` (writes file) `gemfire_properties_path`(returns path) and `gemfire_properties`(returns filename), mirroring the organization of the `create_client_cache` etc.

lib/java_buildpack/container/tomcat/tomcat_geode_store.rb
```diff
@@ -44,8 +44,10 @@ def release
         credentials = @application.services.find_service(FILTER, KEY_LOCATORS, KEY_USERS)['credentials']
         user = credentials[KEY_USERS].find { |u| cluster_operator?(u) }
 
-        @droplet.java_opts.add_system_property 'gemfire.security-username', user['username']
-        @droplet.java_opts.add_system_property 'gemfire.security-password', user['password']
+        File.open(@droplet.root + 'gemfire.properties', 'w') do |f|
```
ekcasey, line 47: Do we ever need to worry about merging with an existing `gemfire.properties` file. E.g do you imagine users would ever configure other features of their app with a `gemfire.properties` in the app root?

ekcasey, line 47: Why `gemfire.properties` instead of `gfsecurity.properties`? https://gemfire.docs.pivotal.io/910/geode/reference/topics/gemfire_properties.html

ekcasey, line 47: How does the session manager find this file. Is it possible to move it to `@droplet.sandbox`?

If it needs to be in the `root` can we create it in the sandbox and symlink it to the root?

lib/java_buildpack/container/tomcat/tomcat_geode_store.rb
sabbey37: Ah, ok, good suggestion!

lib/java_buildpack/container/tomcat/tomcat_geode_store.rb
```diff
@@ -44,8 +44,10 @@ def release
...
+          f.write('security-username=' + user['username'] + "\nsecurity-password=" + user['password'])
+        end
+
         @droplet.java_opts.add_system_property 'gemfire.security-client-auth-init',
```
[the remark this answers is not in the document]
sabbey37: We could. I'm not sure if there's any benefit to seeing that method name printed out in the `JAVA_OPTS`, but I'll discuss it with Ben.

lib/java_buildpack/container/tomcat/tomcat_geode_store.rb
sabbey37: We did think about using `gfsecurity.properties`, but I don't think we've revisited that since we got everything functioning properly.  It would probably be better to use the `gfsecurity.properties` file, though.  I'll update everything.

sabbey37: This is a good point.  It's something Ben and I discussed in the process, but have not revisited.  I will do some experimentation.

sabbey37: The properties file is picked up in the current directory, home directory, or the CLASSPATH (https://geode.apache.org/docs/guide/12/configuring/running/default_file_specs.html).  We tried a few locations, but were only ever able to get it to pick up in the `@droplet.root`.  I'm not sure if we ever tried the `sandbox` directly though.  I know we tried it in `sandbox/conf`.  Let me double check what all those directories are and experiment some more.  Is the `@droplet.sandbox` the ideal location for the file?

ekcasey: Maybe using `gfsecurity.properties` instead of `gemfire.properties` mitigates the risk here? Users really shouldn't be pushing a sensitive credentials in their app.

lib/java_buildpack/container/tomcat/tomcat_geode_store.rb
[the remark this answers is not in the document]
sabbey37: Ben and I discussed it, we think it might be useful to leave this in the JAVA_OPTS in case anyone needs it for debugging purposes.  Also, I guess it makes the most sense now that we've removed the `gemfire.properties` file.

lib/java_buildpack/container/tomcat/tomcat_geode_store.rb
sabbey37: Deleted the addition of this file altogether.

lib/java_buildpack/container/tomcat/tomcat_geode_store.rb
sabbey37: Deleted the addition of this file altogether in favor of pulling the credentials directly from `VCAP_SERVICES` in our ClientAuth class.

sabbey37: Deleted the addition of this file altogether in favor of pulling the credentials directly from `VCAP_SERVICES` in our ClientAuth class.

sabbey37: Deleted the addition of this file altogether in favor of pulling the credentials directly from `VCAP_SERVICES` in our ClientAuth class.

ekcasey (approved):

Issue #848, opened 2021-01-27
Multiple JREs in the offline package


ghaug:
Hi,

we are using the offline package of the Java Buildpack and have the problem that we need to bundle more than one JRE with it. We can't uncomment more than one JRE in the `components.yml` file and a download on deployment is not possible (disabled and no network anyway). We have a little hack in place that additionally bundles SapMachine JRE so our customers can select it if they wish. However, we would prefer a more elegant solution. 

I've made a tiny change that reads `ADD_TO_CACHE` during the bundling process and adds components to the offline cache that are not  specified in `components.yml`. Would there be any interest from your side that I create a pull request?

Thanks and best regards,
Gunter

ekcasey:
@ghaug Thanks! This sounds like a feature that would be broadly useful to `java-buildpack` consumers. I wonder, would it be more intuitive to simply allow multiple JREs in `components.yml` to be uncommented?

ghaug:
Yes, we could allow for multiple JREs in `components.yml` as well. However, we would need to specify a kind of a default JRE because manifests may not specify a JRE. Defining this simply by the order in `components.yml` is probably not a good idea. That's why I thought about this simple solution.

ghaug:
Anyway, as there is general interest, I'll prepare a pull request so we have something to discuss about.

ghaug:
I finally succeeded in creating this [pull request](https://github.com/cloudfoundry/java-buildpack/pull/852). Sorry for the delay.

RealCLanger:
@ekcasey Is there any ETA for when somebody could have a look at this pull request resp. merge it? Thanks a lot in advance!

dmikusa-pivotal:
The PR has been closed, so I'm going to close this issue. Feel free to reopen if there are still outstanding needs here.

PR #852, opened 2021-02-02
Allow for additional components (JREs) in the cache

ghaug:
As described in this [issue](https://github.com/cloudfoundry/java-buildpack/issues/848) we are using the offline package of the Java Buildpack and have the problem that we need to bundle more than one JRE with it. We can't uncomment more than one JRE in the components.yml file and a download on deployment is not possible (disabled and no network anyway). We have a little hack in place that additionally bundles SapMachine JRE so our customers can select it if they wish. However, we would prefer a more elegant solution.

I've made a tiny change that reads ADD_TO_CACHE during the bundling process and adds components to the offline cache that are not specified in components.yml

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,9 +158,11 @@ The offline package is a version of the buildpack designed to run without access
 
 To pin the version of dependencies used by the buildpack to the ones currently resolvable use the `PINNED=true` argument. This will update the [`config/` directory][] to contain exact version of each dependency instead of version ranges.
 
+Only packages referenced in the [`config/components.yml` file][] will be cached. Additional packages may be added using the `ADD_TO_CACHE` argument. It has to be set to the name of a `.yml` file in the [`config/` directory][]). Multiple file names may be separated by colons. This is useful to add additional JREs.
+
 ```bash
 $ bundle install
-$ bundle exec rake clean package OFFLINE=true PINNED=true
+$ bundle exec rake clean package OFFLINE=true PINNED=true ADD_TO_CACHE=sap_machine_jre
 ...
 Creating build/java-buildpack-offline-cfd6b17.zip
 ```
--- a/rakelib/dependency_cache_task.rb
+++ b/rakelib/dependency_cache_task.rb
@@ -109,7 +109,10 @@ module Package
     end
 
     def component_ids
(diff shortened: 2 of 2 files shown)


RealCLanger (approved):
This looks good to me. Thanks for doing it. @maintainers: Is there any chance to get this merged?

ekcasey:
Hey @ghaug thanks for the contribution and apologies for the slow response.

A couple things:

1. This now fails if you don't specify `ADD_TO_CACHE`:
```
bundle exec rake clean package OFFLINE=true PINNED=true
rake aborted!
NoMethodError: undefined method `split' for nil:NilClass
/Users/ekcasey/workspace/cloudfoundry/java-buildpack/rakelib/dependency_cache_task.rb:113:in `component_ids'
/Users/ekcasey/workspace/cloudfoundry/java-buildpack/rakelib/dependency_cache_task.rb:45:in `initialize'
/Users/ekcasey/workspace/cloudfoundry/java-buildpack/Rakefile:42:in `new'
/Users/ekcasey/workspace/cloudfoundry/java-buildpack/Rakefile:42:in `<top (required)>'
/Users/ekcasey/.rbenv/versions/2.5.8/bin/bundle:23:in `load'
/Users/ekcasey/.rbenv/versions/2.5.8/bin/bundle:23:in `<main>'
(See full trace by running task with --trace)
```

2. After Digging into this more, I suspect the feature we are trying to enable already works, if you uncomment multiple JREs in [`components.yml`](https://github.com/cloudfoundry/java-buildpack/blob/f888b34fbe93671ea74c003bea91cda8bd4bb21a/config/components.yml#L31). The [first JRE will be the default](https://github.com/cloudfoundry/java-buildpack/blob/main/lib/java_buildpack/buildpack.rb#L67-L71) and others can be set with `JBP_CONFIG_COMPONENTS`. Maybe we should just document the existing behavior instead of adding the `ADD_TO_CACHE` env var?

ghaug:
Hey @ekcasey thanks for looking into this.

re your comments:

1. Sorry for the stupid mistake, I'll fix that.
2. Has this feature been recently introduced? When I tried it didn't work like that.

ghaug:
I've updated the PR to fix the bug @ekcasey mentioned as 1.

ekcasey (approved):

PR #853, opened 2021-02-03
Detect Geode Tomcat module version

sabbey37:
Previously, we had a hardcoded session manager class name in the `tomcat_geode_store.rb`.  This commit would search through all the jars to find the Geode Tomcat module jar.  It would extract the version of the Geode Tomcat module from the jar name.  It then constructs the session manager class name with the correct Geode Tomcat version.  This is beneficial for users that might want to use a newer version of the Java buildpack, but an older version of the `geode_store` tar that might have a different Geode Tomcat module jar.

Considerations for reviewers:
- What should happen if two Geode Tomcat module jars are found?  EDIT: we changed this so a Runtime Error is thrown if more than one Geode Tomcat module jar is found. Realistically, there should never be more than one and having more than one may cause other issues.
- What should happen if no `geode_tomcat_version` is extracted from the jars?  Currently, we set it to `9` if this happens.
- This line:
```
puts "       Detected Geode Tomcat#{geode_tomcat_version} module"
```
results in this getting printed out in the terminal after a `cf push` (alongside other information):
```
 -----> Downloading geode-store-0.0.2.tar.gz 0.0.2 from https://java-buildpack-tomcat-gemfire-store.s3-us-west-2.amazonaws.com/geode-store-0.0.2.tar (found in cache)
          Expanding geode-store-0.0.2.tar.gz to .java-buildpack/tomcat/lib (0.0s)
          Detected Geode Tomcat8 module
          Adding Geode-based Session Replication
```
Is that ok? Do we want to log something different or not log anything?


lib/java_buildpack/container/tomcat/tomcat_geode_store.rb
```diff
@@ -133,6 +133,24 @@ def create_cache_client_xml
...
+
+            geode_tomcat_version = geode_modules_tomcat_pattern.match(file).named_captures['version']
+          end
+        end
+
+        geode_tomcat_version = '9' if geode_tomcat_version.nil? || geode_tomcat_version.empty?
+        puts "       Detected Geode Tomcat#{geode_tomcat_version} module"
+        @session_manager_classname =
+          "org.apache.geode.modules.session.catalina.Tomcat#{geode_tomcat_version}DeltaSessionManager"
+      end
```
ekcasey, line 152: Can you explain the compatibility matrix to me? I got the impression from the docs that the correct version of the session manager classname depends on the version of Tomcat. Does this mean that if I am using Tomcat 8 I can't use `geode-store `1.13.0`?

ekcasey: Also the [docs](https://github.com/cloudfoundry/java-buildpack/blob/main/docs/container-tomcat.md#configuration) don't seem to indicate the `geode_store.version` is configurable, unlike `redis_store.version` for example. Given that it appears to be configurable, can we add that information to the docs ad part of this PR, including compatibility gotchas if there are any?

sabbey37: It basically just depends on the Geode Tomcat module jar that's included in the tar.  So we could (or a customer could) create a 1.13.0 tar that includes Tomcat 8 and use that in the buildpack. Right now though, of the tars that we have created for customers, compatibility is the following:
0.0.2 = Tomcat8
1.11.0 - 1.13.0 = Tomcat9

We wanted the version to be auto detected from the jars so that if a customer is using a newer Java buildpack, but wants to change their geode store version (from 1.11.0 to 0.0.2, for example), they don't also have to change this class name.

ekcasey: Are [these docs](https://geode.apache.org/docs/guide/19/tools_modules/http_session_mgmt/tomcat_setting_up_the_module.html) wrong then where they say:
> **Depending on the version of Tomcat you are using**, add one of the following lines to $CATALINA_HOME$/conf/context.xml within the <Context> tag:

(emphasis mine)

sabbey37: @ekcasey Those documents are right.  However, the Java buildpack is creating the `context.xml` for the client here, so this is where the user would have to update that version.  We wanted it to be detected automatically so they didn't have to update it.

ekcasey: Let me explain what I think I understand so far and you can tell me what I am missing:
1. Right now users can modify the buildpack to configure the tomcat version.
1. ~Users cannot configure the version of geode store~
1. Users [can](https://github.com/cloudfoundry/java-buildpack/blob/main/config/tomcat.yml#L42) modify the buildpack to configure the version of geode store but it is [not documented](https://github.com/cloudfoundry/java-buildpack/blob/main/docs/container-tomcat.md#configuration)
1. At the point we are writing `context.xml` the tomcat version is known.
1. `geode-store-x.tar` contains one of `Tomcat8DeltaSessionManager` or `Tomcat9DeltaSessionManager` but not both
1. `Tomcat9DeltaSessionManager` will not work with tomcat 8 and vice versa

So we can pick the right class name based on:
1. What exists in geode store, but it could be potentially incompatible with the tomcat version, depending on how users have configured the buildpack.
2. The version of tomcat but it could not exist in the provided version of geode store depending on how users have configured the buildpack.

But either way there could be potentially incompatibility, depending on how users have configured the buildpack.

Not sure that option 1 or 2 is better. You update (option 1) is certainly better than the hardcoded class name we had before. But maybe we should:
1. fail if the versions are incompatible?
2. document the fact that you can configure geode store (only tangentially related)

I guess I slightly prefer choosing the classname based on the tomcat version, because someday you might want to ship a geode store that is compatible with both tomcat 8 & 9?

sabbey37: Some tomcat versions are compatible and others are now.  8 and 9 seem to be compatible, but neither are compatible with 7.  I don't know about 10 yet.  We do plan on making a Tomcat 10 module though.  My ideal state (which I'm not sure we'll be able to achieve due to changing resource allocation) is to create a generic session manager class (like Redis does).

We updated the PR to check the geode-modules-tomcat version we have in the geode_store tar against the buildpack's Tomcat version.  We print a warning to the user if they do not match, but do not fail in case they are otherwise compatible (like 8 and 9).  We also left the option for an empty version number in the future in case we do reach that ideal state of a generic class/module.

ekcasey (approved):

PR #854, opened 2021-02-10
Updated to pull the latest Agent.

jy2697:
Removed the major version, so it always pull the latest Agent from the repository.


ekcasey:
@jy2697 Would you mind signing the CLA? (see link in the comment above)

PR #856, opened 2021-02-18
Update geode store version

sabbey37:
Stick with pulling `1.11.0` version of geode-store to avoid conflicts with a TGF4VMs 1.11.0 tile when future (1.11.+) `.tar` files are created.

### DIFF
--- a/config/tomcat.yml
+++ b/config/tomcat.yml
@@ -40,5 +40,5 @@ redis_store:
   timeout: 2000
   connection_pool_size: 2
 geode_store:
-  version: 1.11.+
+  version: 1.11.0
   repository_root: https://java-buildpack-tomcat-gemfire-store.s3-us-west-2.amazonaws.com
--- a/config/tomcat.yml
+++ b/config/tomcat.yml
@@ -40,7 +40,7 @@ redis_store:
   timeout: 2000
   connection_pool_size: 2
 geode_store:
-  # The version of Geode Store must be less than or equal to the Geode server version to ensure compatibility.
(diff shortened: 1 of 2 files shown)


config/tomcat.yml
```diff
@@ -40,5 +40,5 @@ redis_store:
   timeout: 2000
   connection_pool_size: 2
 geode_store:
-  version: 1.11.+
+  version: 1.11.0
```
ekcasey, line 43: Do you mind adding a comment above this line that explains that patch versions of the client can be incompatible with certain server versions? It's a bit unexpected and I imagine in the future we will forget why we did this and be tempted to change it back.

sabbey37: Sure! After discussing this question with veteran members of the Geode community, I was informed that it is possible for a newer patch version of the Geode client to not be compatible with an older patch version of the Geode server (if a protocol is changed, for example).  We'd like to download the earliest supported version of geode store to make sure we are compatible with all supported TGF4VMs versions.

ekcasey, line 43:
```suggestion
  # The version of Geode Store must be less than or equal to the Geode server version to ensure compatibility.
  #  version is pinned to 1.11.0 to match the oldest supported version of Tanzu Gemfire for VMs
  version: 1.11.0
```
Sorry, I meant this kind of comment.

sabbey37: Yeah, I was thinking about that this morning and realized that's probably what you meant 😅

sabbey37: Just updated it, but let me run the comment by TGF4VMs PMs before merging.

sabbey37: Done! Let us know what you think.

ekcasey (approved):

PR #860, opened 2021-03-16
Implement Sealights Agent component

nadavye:


nadavye:
Hello,
I was wondering if you had some time to look in this PR?

nadavye:
Hi @dmikusa-pivotal ,
I'm reaching out to you as I so you were merging the Datadog PR. I'll appreciate if you could help with reviewing this PR as well.
Thanks,
Nadav

dmikusa-pivotal:
@nadavye Sorry for the delay & thanks for this PR. I'd be happy to review it. I will try to have some feedback for you by the end of the week.

Quick question for you, do you have instructions you can share for testing? Is there a trial account for the service I can use? I like to do an end-to-end acceptance style test before merging a PR like this. Thanks!

lib/java_buildpack/framework/sealights_agent.rb
```diff
@@ -0,0 +1,168 @@
...
+require 'net/http'
+require 'pathname'
+require 'base64'
+require 'json'
+
+module JavaBuildpack
+  module Framework
+
+    # Encapsulates the functionality for enabling zero-touch Sealights support.
+    class SealightsAgent < JavaBuildpack::Component::BaseComponent
```
dmikusa-pivotal, line 30: It is much preferred to use the VersionedDependencyComponent as a base. It will handle all of the download and caching logic for you, something that is not trivial to re-implement. 

https://github.com/cloudfoundry/java-buildpack/blob/main/docs/extending-versioned_dependency_component.md

Is there a specific reason you opted to implement base? Like did VersionedDependencyComponent not work for some reason?

lib/java_buildpack/framework/sealights_agent.rb
```diff
@@ -0,0 +1,168 @@
...
+      end
+
+      # (see JavaBuildpack::Component::BaseComponent#detect)
+      def detect
+        supports? ? "#{self.class.to_s.dash_case}=latest" : nil
+      end
+
+      # (see JavaBuildpack::Component::BaseComponent#compile)
+      def compile
+        announce("'compile' method is starting")
```
dmikusa-pivotal, line 44: There's quite a bit of debug logging like this entry here. Please remove that for your final code submission.

lib/java_buildpack/framework/sealights_agent.rb
```diff
@@ -0,0 +1,168 @@
...
+        download_url = get_download_url
+        download_uri(download_url)
+        extract_zip('sealights-java.zip', get_agent_path)
+        announce("'compile' method has ended")
+      end
+
+      def get_download_url
```
dmikusa-pivotal, line 51: Can you explain how you typically download the agent for Sealights? It seems like the bound service would contain a URL and the agent can be obtained through that URL. Is it possible to download the agent in advance or is the downloaded agent customized in some way/unique to the individual downloading it? 

Also, how would this work in an environment where there is no access to the Internet? Is it possible for Sealights to be hosted on the LAN, so that the server you'd be taking to here would also be on the LAN? Or is the Internet a requirement for the product? 

Thanks for clarifying

lib/java_buildpack/framework/sealights_agent.rb
```diff
@@ -0,0 +1,168 @@
...
+
+      protected
+
+      def download_uri(full_url)
+        announce("Downloading Sealights Agent from '#{full_url}'")
+        begin
+          credentials = @application.services.find_service(FILTER, TOKEN)['credentials']
+          token = credentials[TOKEN]
+          uri = URI.parse(full_url)
+          ENV['http_proxy'] = 'http://127.0.0.1:8888' if credentials.key? 'proxy'
```
dmikusa-pivotal, line 118: Can you explain the intent here? Typically a user would set http_proxy & https_proxy as part of their environment variables which would be set when the buildpack runs. This means as a buildpack, you basically just need to acknowledge those environment variables. You don't need to have a specific configuration for proxy.

lib/java_buildpack/framework/sealights_agent.rb
```diff
@@ -0,0 +1,168 @@
...
+
+          response = Net::HTTP.start(uri.host, uri.port,
+                                     use_ssl: uri.scheme == 'https',
+                                     verify_mode: OpenSSL::SSL::VERIFY_NONE) do |http|
+            request = Net::HTTP::Get.new uri
+            auth = "Bearer #{token}"
+            @logger.info { auth }
```
dmikusa-pivotal, line 125: We definitely cannot log this. It looks like it would contain the user's bearer token, so that would leak credentials into the log stream. Please remove this.

spec/java_buildpack/framework/sealights_agent_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+    end
+
+    # eyJhbGciOiJSUzUxMiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJodHRwczovL1BST0QtU1RBR0lORy5hdXRoLnNlYWxpZ2h0cy5pby8iLCJqd3RpZCI6IlBST0QtU1RBR0lORyxpLTAxM2EyZTU1N2YzOWUzMzNkLEFQSUdXLTE2NGE0NTJkLTY1ZTktNGY4ZC1iODVkLWRjNWJlNGJhZDM2NSwxNDg2MDQ0MjI1NDQ2Iiwic3ViamVjdCI6IlNlYUxpZ2h0c0BhZ2VudCIsImF1ZGllbmNlIjpbImFnZW50cyJdLCJ4LXNsLXJvbGUiOiJhZ2VudCIsIngtc2wtc2VydmVyIjoiaHR0cHM6Ly9QUk9ELVNUQUdJTkctZ3cuc2VhbGlnaHRzLmNvL2FwaSIsInNsX2ltcGVyX3N1YmplY3QiOiIiLCJpYXQiOjE0ODYwNDQyMjV9.O5YrJTiwCj7Nd_5A9-3TGz1Rv_MEj47S6QRpTkfGRwokOiMNh15XybiZ_yOmhbEUYSmNFos84tIlPyrQbySqI7XrksgSWxbM-bysbnXRwzMAf-z4jcRI5xa-YjhNS5VJ01-CynJrBgCl7htZ5BuyPVCbeEgCWWAsJtApIPlYZCRwGo7DEVc7MZbfBa5Qd0S7RzCukuVX6J-mWWM3Fan6zKAOqCKWRqsDNvzi3kTLAH_7-ps1MzQLGFhUE6fR-5Z1P1DRFdoi6uHhg-DA2Qo_9alH6Aa78GadR0MnSMLieeMUZxn13zFCjwlVgprkJkkbMZUdEOCVulDe-CKSDWH3bsUhq93gxR1D-WjwT28e9CIYV7BycQqFmAunrXOdBfwQTFQ2YjjVRxdZduqgTaeSdlLC7ZhRScII5K2exseu-YvAd7JHMEyTKIXrz6yDLk-V1kQ1her1o9sqSe9zjIEDXeLxvJhdeo8sfG48fIr4BaFXBvKhUv8Mu6IHvHT10FDZfqR9SuQsGQ8biSMEf-7a7X7tE02pSgjOjYytis3EFc5c4ArEyLRsfNWXNN4FqUh5njon4SYEOZzlIaGu0zdrPuEK0aD4bxvS2cztHHvkBk5E5KTFxGVx8iA-ho8P5EfRrTHYylr7HWvApf3BmaahQjAfv-qt_NxGQvMshdE-D8U
```
dmikusa-pivotal, line 46: What is this magic string?  :)

dmikusa-pivotal (changes requested):

lib/java_buildpack/framework/sealights_agent.rb
nadavye: Yes, I had to set a token in order to download the agent and wasn't sure how to do that.

lib/java_buildpack/framework/sealights_agent.rb
nadavye: Outside of PCF, our customers either downloading the agent using `wget` with token or just make sure that it will be present next to their application. 

Per your question, whether the agent can be downloaded in advance if by advance you mean to be part of the buildpack then yes, also the preferred way is to download it as it allows us to upgrade/downgrade per customer.

Regarding access to the internet, we need to send data to our cloud, thus that's mandatory. That being, we have one prospect that mentioned that he can send data to the internet but can't download anything during the buildpack due to their policy. For this latter case, I'm not sure what's the preferred way of solving that? Should I include a default agent as part of the buildpack?

lib/java_buildpack/framework/sealights_agent.rb
nadavye: The purpose of this row was to verify that even calls being made by the buildpack will use the organizational proxy. I must say that I have 0 experience in PCF, so it is possible that I've introduced something which isn't a best practice.

nadavye:
Thanks, @dmikusa-pivotal - much appreciated. I'll start working on that in the next few days. 
If by any means you think that I'm off track or want to clarify something, I would be more than happy to have a zoom session with you.
🙏

PR #867, opened 2021-04-08
Add support for Datadog APM Javaagent

### LINKED ISSUE #826: Support for DataDog Agent  as part of  java buildpack
It is observed that, DataAgent isnt part of the java buildpack. It would be great if we have support for datadog agent as well.

tylerbenson:
Support for activating Datadog APM auto instrumentation.

The [Datadog Trace Agent](https://github.com/DataDog/datadog-cloudfoundry-buildpack) is not configured as a service binding, so we rely on detecting environment variable config for activation.

Closes #826

### DIFF
--- a/lib/java_buildpack/framework/datadog_javaagent.rb
+++ b/lib/java_buildpack/framework/datadog_javaagent.rb
@@ -1,0 +1,58 @@
+# frozen_string_literal: true
+
+# Cloud Foundry Java Buildpack
+# Copyright 2013-2021 the original author or authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require 'java_buildpack/component/versioned_dependency_component'
+require 'java_buildpack/framework'
+
+module JavaBuildpack
+  module Framework
+
+    # Encapsulates the functionality for enabling zero-touch Elastic APM support.
+    class DatadogJavaagent < JavaBuildpack::Component::VersionedDependencyComponent
+      include JavaBuildpack::Util
+
+      # (see JavaBuildpack::Component::BaseComponent#compile)
+      def compile
+        download_jar
+      end
+
+      # (see JavaBuildpack::Component::BaseComponent#release)
+      def release
+        java_opts   = @droplet.java_opts
+        java_opts.add_javaagent(@droplet.sandbox + jar_name)
+
+        if !@application.environment.key?('DD_SERVICE')
+          java_opts.add_system_property('dd.service', @application.details['application_name'])
+        end
+
+        if @application.details['application_version']
+          java_opts.add_system_property('dd.version', @application.details['application_version'])
+        end
+      end
+
+      protected
+
+      # (see JavaBuildpack::Component::VersionedDependencyComponent#supports?)
+      def supports?
+        api_key_defined = @application.environment.key?('DD_API_KEY') && !@application.environment['DD_API_KEY'].empty?
+        apm_disabled = @application.environment['DD_APM_ENABLED'] == 'false'
+        apm_enabled = @application.environment['DD_APM_ENABLED'] == 'true'
+        (api_key_defined && !apm_disabled) || apm_enabled
+      end
+    end
+  end
+end
--- a/config/components.yml
+++ b/config/components.yml
@@ -49,6 +49,7 @@ frameworks:
   - "JavaBuildpack::Framework::ContainerCustomizer"
   - "JavaBuildpack::Framework::ContainerSecurityProvider"
   - "JavaBuildpack::Framework::ContrastSecurityAgent"
+  - "JavaBuildpack::Framework::DatadogJavaagent"
   - "JavaBuildpack::Framework::Debug"
   - "JavaBuildpack::Framework::DynatraceAppmonAgent"
   - "JavaBuildpack::Framework::DynatraceOneAgent"
--- a/config/datadog_javaagent.yml
+++ b/config/datadog_javaagent.yml
@@ -1,0 +1,19 @@
+# Cloud Foundry Java Buildpack
+# Copyright 2013-2021 the original author or authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Configuration for the Datadog APM Javaagent
+---
+version: +
+repository_root: https://raw.githubusercontent.com/datadog/dd-trace-java/cloudfoundry/
--- a/docs/framework-datadog_javaagent.md
+++ b/docs/framework-datadog_javaagent.md
@@ -1,0 +1,40 @@
+# Datadog APM Javaagent Framework
+The [Datadog APM]() Javaagent Framework allows your application to be dynamically instrumented [by][datadog-javaagent] `dd-java-agent.jar`.
+
+<table>
+  <tr>
+    <td><strong>Detection Criterion</strong></td><td>One of the following environment variables configured:
+      <ul>
+        <li><code>DD_APM_ENABLED</code> configured as <code>true</code></li>
+        <li><code>DD_API_KEY</code> defined with the assumption of a <a href='https://github.com/DataDog/datadog-cloudfoundry-buildpack'>datadog-cloudfoundry-buildpack</a> configured, and <code>DD_APM_ENABLED</code> not <code>false</code></li>
+      </ul>
+    </td>
+  </tr>
+  <tr>
+    <td><strong>Tags</strong></td>
+    <td><tt>datadog-javaagent=&lt;version&gt;</tt></td>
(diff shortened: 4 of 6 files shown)


tylerbenson:
(Working on the CLA...)

lpriima (approved):

dmikusa-pivotal:
Thanks for the submission! I will take a look and get back to you with feedback shortly.

devinsba:
@dmikusa-pivotal do you know if there is some way for us to influence the memory calculator within this plugin? We need to add to the class count, but not all classes in our jar would be picked up by the counting logic here: https://github.com/cloudfoundry/java-buildpack/blob/main/lib/java_buildpack/jre/open_jdk_like_memory_calculator.rb#L75-L90 because we hide them from the default classloader by changing the file extensions

dmikusa-pivotal:
I don't know of an easy way to do that off-hand. You're adding a component to the frameworks list, which I believe gets processed after the JRE components run. That means it would have already made its decision about how many classes are present.

Do you have a rough idea of how many classes you're talking about? My gut feeling is that unless it's a lot, it probably won't matter. The metaspace size allocated by the memory calculator is just an educated guess, it's not possible to get a perfect calculation. If there were a case where a user did enable your agent and it caused issues, perhaps if they were trying to run with a small amount of memory, they could adjust the class count manually or add headroom.

dmikusa-pivotal:
Looked a bit more and I don't see a way you could do that, not from a framework.

This is the method that basically drives the release process, which is where the command is generated: https://github.com/cloudfoundry/java-buildpack/blob/main/lib/java_buildpack/buildpack.rb#L90-L113

You'll see there that only JRES & Containers (like Tomcat) get to contribute to the start command. Frameworks do not. Nor is there a way to modify the command, it's generated in this method and only available there.

devinsba:
> Do you have a rough idea of how many classes you're talking about? My gut feeling is that unless it's a lot, it probably won't matter. The metaspace size allocated by the memory calculator is just an educated guess, it's not possible to get a perfect calculation. If there were a case where a user did enable your agent and it caused issues, perhaps if they were trying to run with a small amount of memory, they could adjust the class count manually or add headroom.

We have had a customer hit OOM issues when using this buildpack with our agent added manually. We are adding between 2k and 3k classes to the classpath, most of which can't be found by the classpath scanning tool because they have been hidden from java's default classloader intentionally (by changing the file extensions)

Maybe we could add another configuration flag to add an offset on the memory calculator before it is run? That's all I can think of. I think the user we are talking to could add that

dmikusa-pivotal:
>hidden from java's default classloader intentionally (by changing the file extensions)

What do these hidden classes look like? Perhaps we could modify the method we use to count classes to look for these as well?

tylerbenson:
@dmikusa-pivotal how does the memory calculator work for other framework provided javaagents? I don't see how it takes them into account...

Our javaagent "hides" the internal classes by renaming the file to `.classdata` instead of `.class`. This is an implementation detail that doesn't seem like it would make sense to handle. I think it would be better if the framework could influence the total somehow.

Do you see any issues with the PR otherwise?

dmikusa-pivotal:
>@dmikusa-pivotal how does the memory calculator work for other framework provided javaagents? I don't see how it takes them into account...

I don't think they do either. I looked through a few and I don't see anything that looks to be trying to influence the memory calculator. While it's possible for a framework to set Java opts (system properties, agents and really any property the JVM accepts), I don't see frameworks doing this with regards to memory. At the time the framework runs, you can't know how much memory the runtime container will have and so you can't make good decisions about this. The memory calculator has to do this and it has to do it just prior to the app starting up.

The only way to do this would be to adjust the values being passed to the memory calculator, but the way that we generate the start command and the arguments passed to the memory calculator doesn't presently allow for you to influence its calculations.

>Our javaagent "hides" the internal classes by renaming the file to .classdata instead of .class. This is an implementation detail that doesn't seem like it would make sense to handle. I think it would be better if the framework could influence the total somehow.

OK, noted. I don't really want to do this, at least not now, but I was curious and I wanted to take a note for future reference.

>Do you see any issues with the PR otherwise?

I looked through it and it looks fine. I do want to give it a try though, in particular, to run through the instructions and test those. Will this work with a DataDog free trial? Or is there anything else I'd need to test it? Thanks

tylerbenson:
It should work with a free trial just fine.  If you run into any problems, please let me know.

dmikusa-pivotal:
## Notes:

- Setting `DD_APM_ENABLED: true` resulted in the agent being installed but not functional (`WARN datadog.trace.agent.common.writer.ddagent.DDAgentApi - Error while sending 3 (size=2KB) traces to the DD agent. Total: 3, Received: 3, Sent: 0, Failed: 3. java.net.ConnectException: Failed to connect to localhost/127.0.0.1:8126 (Will not log errors for 5 minutes)`. I expected this to fail, since there's no key.

- Setting `DD_API_KEY: <key>` resulted in the agent being installed but still not functional, same `WARN` message. I was expecting this to work.

- I see this message in the docs, which is a little confusing to me `DD_API_KEY defined with the assumption of a datadog-cloudfoundry-buildpack configured,`. It sounds like it is a requirement for this feature to also be using the DataDog buildpack, so I added that into my manifest.yml's buildpacks group (pointing to `https://github.com/DataDog/datadog-cloudfoundry-buildpack#4.21.0`). This results in seeing `Installing Datadog IOT Agent, Dogstatsd and Trace Agent` in the output, as well as the Java agent installed by the JBP. There are a couple messages to STDERR though:

```
   cp: cannot stat '/tmp/buildpackdownloads/5db9f9155b0dedb3ec1ccf91597ae2dc/lib/trace-agent': No such file or directory
   chmod: cannot access '/tmp/app/datadog/trace-agent': No such file or directory
```

There are also some issues in the app logs too:

```
   2021-04-16T11:23:09.95-0400 [APP/PROC/WEB/0] OUT Skipping log endpoint validation
   2021-04-16T11:23:09.95-0400 [APP/PROC/WEB/0] OUT starting datadog
   2021-04-16T11:23:09.95-0400 [APP/PROC/WEB/0] OUT ~/datadog ~
   2021-04-16T11:23:10.03-0400 [APP/PROC/WEB/0] OUT ~
   2021-04-16T11:23:10.03-0400 [APP/PROC/WEB/0] ERR .profile.d/02-run-datadog.sh: line 94: ./dogstatsd: No such file or directory
   2021-04-16T11:23:10.04-0400 [APP/PROC/WEB/0] ERR .profile.d/02-run-datadog.sh: line 97: ./trace-agent: No such file or directory
```

and ultimately, I still see the WARN message about not being able to connect to localhost:8126.

- I tried the instructions under [log configuration](https://github.com/DataDog/datadog-cloudfoundry-buildpack#log-collection) for the DD buildpack to see if that might help. I'm still seeing the same core WARN message though. I also see some more messages like `Resetting buildpack log redirection` which is repeating fairly often.

- Ultimately, the dependent supply buildpack doesn't seem to be installing the binaries I need. My guess is because I'm not building and uploading the buildpack, but using it directly from Github. I don't have a CF env where I can build and upload a buildpack at the moment to test though.

## Thoughts:

Largely, I think my problems above are with the CF environment I used. I'll work on getting another one where I have admin control to try again. Ignoring my troubles, the PR here looks pretty good. I do think there are a few things we can clarify though, see below.

1. I think that the docs should clarify the relationship between this feature in the JBP and the DataDog buildpack. It seems to be required to use this other buildpack too, so we should absolutely clarify that in the JBP docs page for DataDog. The doc page should explain everything that is required to make this feature work.
2. I am assuming that the DD_API_KEY is required, but it seems like the detection logic will enable the agent if you simply set DD_APM_ENABLED to true, even if there is no key. What is the intent with DD_APM_ENABLED here? Is it possible for the agent to work without an API key and still do something meaningful? 
3. I'd like to see a section on the JBP docs page for DataDog that lists the supported `DD_*` env variables that can be set or if those are not unique to the buildpack, a link to the DataDog agent docs where these are described.
4. If the DataDog buildpack is required (see thought #1), then I think the feature in the JBP should print a WARNING when it is enabled, but the other buildpack isn't present. For example, if I set DD_APM_ENABLED=true or add a DD_API_KEY, that will enable this functionality in the JBP. At the same time, if the DataDog buildpack isn't present, functionality is not going to work right, so the JBP should log a WARNING and indicate that requirement is not met.

dmikusa-pivotal:
OK, as expected, once I got a CF with admin access and uploaded a packaged DataDog buildpack, it all started working OK. The errors went away.

I think the four thoughts I mentioned still apply. I'd like to get those worked out so the docs are a little more clear and the detection is a little more robust. Then I think we'll be all set. Let me know if you have any questions.

lib/java_buildpack/framework/datadog_javaagent.rb
```diff
@@ -0,0 +1,58 @@
...
+        end
+      end
+
+      protected
+
+      # (see JavaBuildpack::Component::VersionedDependencyComponent#supports?)
+      def supports?
+        api_key_defined = @application.environment.key?('DD_API_KEY') && !@application.environment['DD_API_KEY'].empty?
+        apm_disabled = @application.environment['DD_APM_ENABLED'] == 'false'
+        apm_enabled = @application.environment['DD_APM_ENABLED'] == 'true'
```
zippolyte, line 53: I think this env var should be `DD_APM_CONFIG_ENABLED`

lib/java_buildpack/framework/datadog_javaagent.rb
```diff
@@ -0,0 +1,58 @@
...
+        (api_key_defined && !apm_disabled) || apm_enabled
```
zippolyte, line 54: Why enable when there is no API key ? I'd have just `api_key_defined && apm_enabled`

tehall16:
Hi @tylerbenson @zippolyte, if you can prioritize that would be great as our mutual customer is asking for this ASAP.  We want to release the next JVM and include these changes, but if they aren't made soon it will be another month before the next release.

lib/java_buildpack/framework/datadog_javaagent.rb
dmikusa-pivotal: I changed this.

dmikusa-pivotal:
OK, so I just submitted a change that addresses 1-4 in my previous comment. Basically, API key is now required, it won't add the agent unless the datadog buildpack runs & includes the binaries in the `datadog/` folder, docs mention all this and I found a doc link to use that reference the agent config options available.

I'm going to merge this now so it gets into the next release. If you have more feedback, just open a new PR with changes and I'll take a look. Thanks!

PR #873, opened 2021-05-12
Implement Sealights Agent framework

alonweiss-sl:

### DIFF
--- a/docs/framework-sealights_agent.md
+++ b/docs/framework-sealights_agent.md
@@ -1,0 +1,43 @@
+# Sealights Agent Framework
+The Sealights Agent Framework causes an application to be automatically configured to work with [Sealights Service][].
+
+<table>
+  <tr>
+    <td><strong>Detection Criterion</strong></td><td>Existence of a single bound sealights service. The existence of a sealights service defined by the <a href="http://docs.cloudfoundry.org/devguide/deploy-apps/environment-variable.html#VCAP-SERVICES"><code>VCAP_SERVICES</code></a> payload containing a service name, label or tag with <code>sealights</code> as a substring.
+</td>
+  </tr>
+  <tr>
+    <td><strong>Tags</strong></td><td><tt>sealights-agent=&lt;version&gt;</tt></td>
+  </tr>
+</table>
+Tags are printed to standard output by the buildpack detect script
+
+## User-Provided Service
+When binding Sealights using a user-provided service, it must have name or tag with `sealights` in it.
+The credential payload can contain the following entries. 
+
+| Name | Description
+| ---- | -----------
+| `token` | A Sealights Agent token
+
+## Configuration
+For general information on configuring the buildpack, including how to specify configuration values through environment variables, refer to [Configuration and Extension][].
+
+The framework can be configured by modifying the [`config/sealights_agent.yml`][] file. The framework uses the [`Repository` utility support][repositories] and so it supports the [version syntax][] defined there.
+
+| Name | Description
+| ---- | -----------
+| `build_session_id` | Sealights build session id for the application. Leave blank to use the value embedded in the jar/war artifact
+| `proxy` | Specify a proxy
+| `lab_id` | Specify a Sealights Lab Id
+| `auto_upgrade` | Enable/disable agent auto-upgrade. Off by default
+
+## Troubleshooting
+
+If your container is running out of memory and exited with status 137, then you should setup and use a remote collector as explained in the `User-Provided Service` above section.
+
+[`config/sealights_agent.yml`]: ../config/sealights_agent.yml
+[Configuration and Extension]: ../README.md#configuration-and-extension
+[repositories]: extending-repositories.md
+[version syntax]: extending-repositories.md#version-syntax-and-ordering
+[Sealights Service]: https://www.sealights.io
--- a/docs/framework-sealights_agent.md
+++ b/docs/framework-sealights_agent.md
@@ -27,17 +27,22 @@ The framework can be configured by modifying the [`config/sealights_agent.yml`][
 
 | Name | Description
 | ---- | -----------
-| `build_session_id` | Sealights build session id for the application. Leave blank to use the value embedded in the jar/war artifact
-| `proxy` | Specify a proxy
-| `lab_id` | Specify a Sealights Lab Id
+| `build_session_id` | Sealights [Build Session ID][] for the application. Leave blank to use the value embedded in the jar/war artifacts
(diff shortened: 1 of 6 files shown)


docs/framework-sealights_agent.md
```diff
@@ -0,0 +1,43 @@
...
+## User-Provided Service
+When binding Sealights using a user-provided service, it must have name or tag with `sealights` in it.
+The credential payload can contain the following entries. 
+
+| Name | Description
+| ---- | -----------
+| `token` | A Sealights Agent token
+
+## Configuration
+For general information on configuring the buildpack, including how to specify configuration values through environment variables, refer to [Configuration and Extension][].
```
dmikusa-pivotal, line 24: A couple notes here:

1. For agent-specific settings, like build_session_id & lab_id, if you can link to documentation that explains these, that is helpful to users, especially newer users.
2. There are a couple of conventions around proxy configuration, to make this more automatic for users. First, the buildpack itself respects `http_proxy` and `https_proxy` ([see here](https://github.com/cloudfoundry/java-buildpack/blob/main/docs/extending-repositories.md#proxies)). If these env variables are set, the buildpack automatically uses the defined proxy to download resources. It is very common for users to set these values, when a proxy is required. Second, the JVM has its own proxy settings, `http.proxyHost` and `http.proxyPort`. These are typically set by a user via the `JAVA_OPTS` env variable and would be present when the app runs. Does your agent support any of these existing conventions? If so, it may automatically configure the proxy. If it doesn't, I would suggest that the buildpack look at http_proxy/https_proxy and, if set, automatically configure the proxy settings for your agent. This should remove the step for a user to manually configure proxy settings in a third location, and because it's so common for users that require a proxy to set these already, it'll likely just work for them without any additional thought.
3. You're missing a `version` label. Something like `version` -> `The version of Sealights Agent to use. Candidate versions can be found in [this listing](link-to-your-index.yml-file)` is common.

docs/framework-sealights_agent.md
```diff
@@ -0,0 +1,43 @@
...
+| Name | Description
+| ---- | -----------
+| `build_session_id` | Sealights build session id for the application. Leave blank to use the value embedded in the jar/war artifact
+| `proxy` | Specify a proxy
+| `lab_id` | Specify a Sealights Lab Id
+| `auto_upgrade` | Enable/disable agent auto-upgrade. Off by default
+
+## Troubleshooting
+
+If your container is running out of memory and exited with status 137, then you should setup and use a remote collector as explained in the `User-Provided Service` above section.
```
dmikusa-pivotal, line 37: I don't see where this is referencing. I would suggest linking to the appropriate section or updating this section to have the instructions.

lib/java_buildpack/framework/sealights_agent.rb
```diff
@@ -0,0 +1,87 @@
...
+
+      def initialize(context)
+        super(context)
+        @logger = JavaBuildpack::Logging::LoggerFactory.instance.get_logger SealightsAgent
+      end
+
+      # (see JavaBuildpack::Component::BaseComponent#compile)
+      def compile
+        download_zip(false)
+        @droplet.copy_resources
```
dmikusa-pivotal, line 37: Is this something you envision being useful? I don't see any [resources being added as part of your commit](https://github.com/cloudfoundry/java-buildpack/tree/main/resources). If you don't have any resources to copy, you can likely remove this call.

lib/java_buildpack/framework/sealights_agent.rb
```diff
@@ -0,0 +1,87 @@
...
+      end
+
+      # (see JavaBuildpack::Component::BaseComponent#release)
+      def release
+        @droplet.java_opts.add_javaagent(agent)
+        credentials = @application.services.find_service(FILTER, TOKEN)['credentials']
+        @droplet.java_opts.add_system_property('sl.token', credentials[TOKEN])
+        @droplet.java_opts.add_system_property('sl.tags', 'pivotal_cloud_foundry')
+        add_system_property 'sl.enableUpgrade', ENABLE_UPGRADE
+        add_system_property 'sl.buildSessionId', BUILD_SESSION_ID
```
dmikusa-pivotal, line 47: I don't know if it will matter here, or for any of your configuration properties, but this `add_system_property` method does *not* handle escaping of values. If a value for any of your configuration options could legitimately contain spaces or any characters that would be shell-interpreted, you'd need to shell-escape them first. 

This got us with another agent recently, you can see how to properly escape them here. https://github.com/cloudfoundry/java-buildpack/blob/main/lib/java_buildpack/framework/app_dynamics_agent.rb#L83 Basically just wrap values with `Shellwords` module & it handles the escaping.

dmikusa-pivotal (changes requested):
Looks pretty good, thanks for the submission. I have a couple of docs clarifications/suggestions for you. Let me know if you have any questions.

docs/framework-sealights_agent.md
alonweiss-sl: 1. Added links to the official docs
2. TL;DR: It's intentional. The http_proxy and https_proxy seem to be mostly relevant for the buildpack phase (please correct me if I'm wrong). This configuration is used at runtime to communicate with the Sealights backend service. It may, or may not, be relevant for all other HTTP operations, therefore we also don't use the http.proxyHost and http.proxyPort system properties -- we've seen a lot of cases where the Sealights' proxy was configured and http.proxyHost wasn't. Additionally, we've seen a few cases where the http.proxyHost/Port were updated at runtime, as part of testing (which is when Sealights agents are used), and wasn't cleared afterwards.
3. Added with the correct links

docs/framework-sealights_agent.md
alonweiss-sl: Updated the section with links to our official documentation pages

lib/java_buildpack/framework/sealights_agent.rb
alonweiss-sl: Thanks for pointing this out. While our tokens are URL/base64 encoded and ought to be safe, I've added this just in case.

docs/framework-sealights_agent.md
dmikusa-pivotal: 1. & 3. 👍 

For 2., I'm OK with that. If you intentionally want it to not automatically pick up proxy settings. I just know that users often expect stuff like that to just work. What I will do, so it's clear, is add a note to the doc that specifically says that `proxy` is not automatically  deduced from http_proxy/https_proxy or similar JVM settings.

PR #880, opened 2021-05-26
Don't set dd.version Java opts if DD_VERSION env var is provided

martinpfeifer:
Hello,

currently the Datadog Java agent integration uses the application version generated by [(https://docs.cloudfoundry.org/devguide/deploy-apps/environment-variable.html#VCAP-APPLICATION)](https://docs.cloudfoundry.org/devguide/deploy-apps/environment-variable.html#VCAP-APPLICATION) and this currently is not configurable.
This PR enables one to set a custom version using the `DD_VERSION` environment variable.
Could you please have a look.

Thanks and best regards
Martin Pfeifer


dmikusa-pivotal:
Looks reasonable to me. Is there any way you could appease the CLA bot? That has to happen before I can merge it. Thanks.

dmikusa-pivotal:
Thanks

martinpfeifer:
The CLA is signed and the bot seems happy.

dmikusa-pivotal:
FYI only. 

There are a couple of issues with the logic. This [test is checking if the version is not present](https://github.com/cloudfoundry/java-buildpack/pull/880/files#diff-d1553c9e9ccc4c2de2f7b9b433ed985a4796a76948c4453754836e71361fce8bR119), so it passes. We want to validate that it's present though, so it needs to be `.to`.

Also, [this bit is always true](https://github.com/cloudfoundry/java-buildpack/pull/880/files#diff-147cc19acd8e3446f89b75f550983e4a7ad9904d1516d0944c7b24cb15ace0c5R53) because, at least in the test environment, `@application.details['application_version']` is always set with a value.

I have fixed this, have passing tests & will merge shortly. Thanks for the submission.

PR #890, opened 2021-07-02
Fix class count discrepancy when using DataDog agent.

### LINKED ISSUE #887: Memory Calculations are wrong for Datadog Javaagent
The Datadog Javaagent uses a custom classloader to avoid exposing private classes onto the system classpath.  This includes renaming class files with a custom class suffix (`.classdata`).  How can we make it so the memory calculations are more accurate?

cc @dmikusa-pivotal

dmikusa-pivotal:
Resolves #887.

This fix will go through the DataDog agent file and count the number of `.classdata` hidden classes. It then creates a shadow JAR file with the same number of fake classes in it. The Java buildpack's normal class counting mechanism will now see these classes and pick a correct class count.


amitshah16:
Hello @dmikusa-pivotal  - How do we load this in our foundation? I work with Luke Perry!

PR #915, opened 2021-11-17
fixed case in networkzone query parameter

arthfl:
Little fix in the case for the `networkzone` query parameter.
The z is now an uppercase Z :-)

### DIFF
--- a/lib/java_buildpack/framework/dynatrace_one_agent.rb
+++ b/lib/java_buildpack/framework/dynatrace_one_agent.rb
@@ -115,7 +115,7 @@ module JavaBuildpack
                        '&bitness=64' \
                        "&Api-Token=#{credentials[APITOKEN]}"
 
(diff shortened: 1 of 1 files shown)


dmikusa-pivotal:
Thanks for the fix. Is there any concern here with backward compatibility? Could perhaps some users still need the old URL, while other users need the new URL?

Also, do you know what happens if we send a request with the lowercase `z`? Is that attribute ignored? Will the request 404?

Thanks

arthfl:
@dmikusa-pivotal 
The request with the lowercase `z` never worked anyways. It didn't cause any errors with the deployment, but the feature wasn't used properly because the API just ignored it anyways.

Meaning: there was no old URL with a lowercase `z`, this was just in the wrong case in there from the beginning.

dmikusa-pivotal:
OK, perfect. Thanks for clarifying & thanks for the PR!

PR #926, opened 2022-01-03
Add Luna API shared library to library path for Java 9+

dmikusa-pivotal:
With Java 9, the extension directory functionality was dropped. This means the buildpack has to handle installation different for Java 8 and Java 9+. With Java 8, we add the required files to the extension directory. With Java 9, we add the required JAR file to the classpath, but this was not sufficient to load the required shared library. To fix this, we are adding the Luna API shared library to the LD_LIBRARY_PATH environment variable. The JVM uses this to locate shared libraries, so it can now find the shared library.


pivotal-david-osullivan (approved):
Looks good!

PR #942, opened 2022-05-10
Switch Luna download URLs

dmikusa-pivotal:
The Luna Security Provider site went down. Thales has provided the binaries and we have relocated them to the standard download repository. This may be temporary, it's not clear at this time, but it will get the buildpack back and working again. It has been broken since the site went down.

### DIFF
--- a/config/luna_security_provider.yml
+++ b/config/luna_security_provider.yml
@@ -16,7 +16,7 @@
 # Configuration for the Luna Security Provider framework
 ---
 version: 7.+
-repository_root: https://files.cf-hsm.io/luna-installer
+repository_root: "{default.repository.root}/luna-security-provider"
 ha_logging_enabled: true
 logging_enabled: false
 tcp_keep_alive_enabled: false


leoclee:
@dmikusa-pivotal Would it be possible to make the 6.2.0 version of lunaclient available in the download repository? We were previously accessing it using https://files.cf-hsm.io/luna-installer/lunaclient-min-6.2.0.tar  Thanks!

dmikusa-pivotal:
@leoclee I'm checking into it. I don't have a copy of that file. I'll let you know what I find out from the vendor. Thanks.

leoclee:
@dmikusa-pivotal Thank you, I appreciate your help with this. The sha256 hash of the file that I recovered from the buildpack cache did not match yours. It looks like my copy of lunaclient-min-6.2.0.tar has an additional EULA.pdf in the root of the tar (with a 2019-03-21 timestamp). I have attached it here in case you want to review it (I had to gz the tar file--github wouldn't let me upload it as a tar).

[lunaclient-min-6.2.0.tar.gz](https://github.com/cloudfoundry/java-buildpack/files/8820279/lunaclient-min-6.2.0.tar.gz)

leoclee:
```
219b14130a0e0d0812ef999215b49ba68de7b90692c35be1eaca2ecc183312d4  lunaclient-min-6.2.0.tar
```

leoclee:
@dmikusa-pivotal Thanks!

PR #945, opened 2022-05-11
Remove Redis Store and Add Redisson

dmikusa-pivotal:
The Tomcat container has support for using Redis to persist sessions. It was previously using the Redis Store project for this, however, that project has been end-of-life for a long time. As such, we've removed the usage of Redis Store and swapped in the Redission library instead. It also provides support for Tomcat session storage in Redis. This project is currently maintained and provides a better outlook for users that need to run Tomcat and store sessions in Redis.


dmikusa-pivotal:
I am going to close out this PR. We have hit some issues in testing it. 

Essentially, the Redisson libraries that get installed have a large surface area including some common libraries like [Netty, Jackson, and SnakeYAML](https://github.com/redisson/redisson/wiki/15.-Dependency-list) all of which can conflict with dependencies from a typical Spring Boot user's application. The buildpack with this PR would install Redisson into Tomcat's common class loader, while apps can pull in a different version of those libraries in the webapp class loader, which can create some difficult to fix/debug problems). 💥 

We have explored some options to limit the surface area of Redisson, but have not found a solution in which we are confident/comfortable enough to move forward.

Issue #643, opened 2018-10-19
NoClassDefFoundError: org/apache/commons/dbcp2/BasicDataSource


mortenri:
Hi,

We updated our project  from Spring Boot .5.15.RELEASE to 2.0.6.RELEASE and Spring Framework from 4.3.18.RELEASE to 5.1.1.RELEASE. We are also using javax.persistence (2.0.4.v201112161009), eclipselink (2.5.2)and liquibase (3.5.3). The service is bound to one postgresql services. As soon as I push the application to cloud foundry we are facing the following error:

- NoClassDefFoundError: org/apache/commons/dbcp2/BasicDataSource

Is there any known issue to the java buildpack?

Regards
Morten

nebhale:
There is nothing specific in the buildpack that would cause this issue.  The buildpack delegates to Spring Cloud Connectors, which has strategies for determine which connection pool to create and configuration.  The switch between Spring Boot 1 and Spring Boot 2 could cause something like this as Spring Boot 2 (and Spring 5 by extension) dropped support for some older connection pools that were no longer maintained.

Issue #483, opened 2017-08-21
Offline Version + Introscope


chlamprecht:
Hi,

I'm trying to use introscope in an offline Package (our cf installation has no connection to the internet). But it looks like theres no introscope package included in the cache folder.
So it seems that the user provided service doesn't have an effect when used at an "cf push".

Any idea how to get introscope working?

nebhale:
Introscope support was disabled until [`v4.4`](https://github.com/cloudfoundry/java-buildpack/releases/tag/v4.4) because there was no legal permission to distribute the Introscope agent.  Now that the CA team is hosting those binaries in public and has created a distribution license that allows us to do so, the offline buildpack contains them.

If you would like to use the Introscope agent before `v4.4`, then you’ll need to [host a repository](https://github.com/cloudfoundry/java-buildpack/blob/master/docs/extending-repositories.md) that contains an index and the binaries as well as updating your buildpack to [point at that self-hosted repository](https://github.com/cloudfoundry/java-buildpack/blob/v4.3/config/introscope_agent.yml#L18).

Issue #451, opened 2017-06-23
Resource Exhaustion triggers Thread Dump


nebhale:
Currently, when resource exhaustion is reached, a histogram and an optional heap dump are created.  When exhaustion is reached, a thread dump should be triggered as well.

This issue is to investigate how easily this can be done and is split from the larger #70.

nebhale:
@gberche-orange We've added a thread-dump to the buildpack.  The output looks like [this](https://gist.github.com/nebhale/f74b8363501b3ac2ecfb016fc8869166#file-thread-dump-log-L116).

gberche-orange:
thanks @nebhale and @glyn ! This looks great.

Issue #419, opened 2017-05-03
Unable to download latest AppDynamics Agent


nkatre:
The latest agent version for AppDynamics agent specified in [index.yml](https://packages.appdynamics.com/java/index.yml) file is 4.3.+
The [config](https://github.com/cloudfoundry/java-buildpack/blob/master/config/app_dynamics_agent.yml#L18) file specifies the agent version as 4+. However the agent downloaded is 4.2.15_0 rather than latest 4.3 agent.

Is there something I am missing ?

nebhale:
If I had to guess, I don't believe that the CloudFront distribution has been properly [invalidated](http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/Invalidation.html) as you've updated the index underneath it.  If you run the following command

```
## Request
curl "https://packages.appdynamics.com/java/index.yml" \
     -H "Accept: */*" \
     -H "If-Modified-Since: Thu, 06 Apr 2017 06:57:34 GMT" \
     -H "Accept-Encoding: gzip;q=1.0,deflate;q=0.6,identity;q=0.3" \
     -H "If-None-Match: 00276f92a3f5896bda7bb619aa7531db" \
     -H "User-Agent: Ruby"
```

You'll received a response that only goes to `4.2.15_2`

```
---
4.1.3_1: https://packages.appdynamics.com/java/4.1.3.1/AppServerAgent-4.1.3.1.zip
4.1.7_1: https://packages.appdynamics.com/java/4.1.7.1/AppServerAgent-4.1.7.1.zip
4.1.7_2: https://packages.appdynamics.com/java/4.1.7.2/AppServerAgent-4.1.7.2.zip
4.1.8_1: https://packages.appdynamics.com/java/4.1.8.1/AppServerAgent-4.1.8.1.zip
4.1.8_2: https://packages.appdynamics.com/java/4.1.8.2/AppServerAgent-4.1.8.2.zip
4.1.8_3: https://packages.appdynamics.com/java/4.1.8.3/AppServerAgent-4.1.8.3.zip
4.1.8_4: https://packages.appdynamics.com/java/4.1.8.4/AppServerAgent-4.1.8.4.zip
4.1.8_5: https://packages.appdynamics.com/java/4.1.8.5/AppServerAgent-4.1.8.5.zip
4.1.8_6: https://packages.appdynamics.com/java/4.1.8.6/AppServerAgent-4.1.8.6.zip
[...]

PR #954, opened 2022-06-16
Add Azul Platform Prime to java-buildpack

ushaazul:
modified:   README.md
	modified:   config/components.yml
	new file:   config/zing_jre.yml
	new file:   docs/jre-zing_jre.md
	new file:   lib/java_buildpack/jre/zing_jre.rb

### DIFF
--- a/lib/java_buildpack/jre/open_jdk_like.rb
+++ b/lib/java_buildpack/jre/open_jdk_like.rb
@@ -38,8 +38,10 @@ module JavaBuildpack
       # (see JavaBuildpack::Component::ModularComponent#sub_components)
       def sub_components(context)
         [
+          JvmkillAgent.new(sub_configuration_context(context, 'jvmkill_agent')),
           OpenJDKLikeJre.new(sub_configuration_context(context, 'jre')
                                .merge(component_name: self.class.to_s.space_case)),
+          OpenJDKLikeMemoryCalculator.new(sub_configuration_context(context, 'memory_calculator')),
           OpenJDKLikeSecurityProviders.new(context)
         ]
       end
--- a/lib/java_buildpack/jre/open_jdk_like.rb
+++ b/lib/java_buildpack/jre/open_jdk_like.rb
@@ -38,10 +38,8 @@ module JavaBuildpack
       # (see JavaBuildpack::Component::ModularComponent#sub_components)
       def sub_components(context)
         [
-          JvmkillAgent.new(sub_configuration_context(context, 'jvmkill_agent')),
           OpenJDKLikeJre.new(sub_configuration_context(context, 'jre')
                                .merge(component_name: self.class.to_s.space_case)),
-          OpenJDKLikeMemoryCalculator.new(sub_configuration_context(context, 'memory_calculator')),
           OpenJDKLikeSecurityProviders.new(context)
         ]
       end
--- a/README.md
+++ b/README.md
@@ -115,6 +115,7 @@ The buildpack supports extension through the use of Git repository forking. The 
   * [YourKit Profiler](docs/framework-your_kit_profiler.md) ([Configuration](docs/framework-your_kit_profiler.md#configuration))
 * Standard JREs
   * [Azul Zulu](docs/jre-zulu_jre.md) ([Configuration](docs/jre-zulu_jre.md#configuration))
+  * [Azul Platform Prime](docs/jre-zing_jre.md) ([Configuration](docs/jre-zing_jre.md#configuration))
   * [GraalVM](docs/jre-graal_vm_jre.md) ([Configuration](docs/jre-graal_vm_jre.md#configuration))
   * [IBM® SDK, Java™ Technology Edition](docs/jre-ibm_jre.md) ([Configuration](docs/jre-ibm_jre.md#configuration))
   * [OpenJDK](docs/jre-open_jdk_jre.md) ([Configuration](docs/jre-open_jdk_jre.md#configuration))
--- a/config/components.yml
+++ b/config/components.yml
@@ -34,6 +34,7 @@ jres:
   - "JavaBuildpack::Jre::OpenJdkJRE"
 # - "JavaBuildpack::Jre::OracleJRE"
 # - "JavaBuildpack::Jre::ZuluJRE"
+# - "JavaBuildpack::Jre::ZingJRE"
 # - "JavaBuildpack::Jre::SapMachineJRE"
 
 # Frameworks are processed in order.
--- a/config/zing_jre.yml
+++ b/config/zing_jre.yml
@@ -1,0 +1,33 @@
+# Cloud Foundry Java Buildpack
+# Copyright 2013-2020 the original author or authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Configuration for JRE repositories keyed by vendor
+# Pre Java 1.8, permgen was used instead of metaspace.  Please see the documentation for more detail.
+
+# You must specify a the repository root of a Azul Platform Prime JRE repository. Please see the documentation for more detail.
+# e.g.  repository_root: "https://example.com/zing-jre/{platform}/{architecture}"
+---
+jre:
+  version: 1.8.0_+
+  repository_root: "https://cdn.azul.com/zing/bin"
+jvmkill_agent:
+  version: 1.+
+  repository_root: "{default.repository.root}/jvmkill/{platform}/{architecture}"
+memory_calculator:
+  version: 3.+
+  repository_root: "{default.repository.root}/memory-calculator/{platform}/{architecture}"
+  class_count: 
+  headroom: 
+  stack_threads: 250
--- a/config/zing_jre.yml
+++ b/config/zing_jre.yml
@@ -21,7 +21,7 @@
 ---
 jre:
   version: 1.8.0_+
-  repository_root: "https://cdn.azul.com/zing/bin"
+  repository_root: ""
 jvmkill_agent:
   version: 1.+
   repository_root: "{default.repository.root}/jvmkill/{platform}/{architecture}"
--- a/docs/jre-zing_jre.md
+++ b/docs/jre-zing_jre.md
@@ -49,34 +49,7 @@ To add the JCE Unlimited Strength `local_policy.jar`, add your file to `resource
 To add custom SSL certificates, add your `cacerts` file to `resources/zing_jre/lib/security/cacerts`.  This file will be overlayed onto the Azul Platform Prime distribution.
 
 ### `jvmkill`
-The `jvmkill` agent runs when an application has experience a resource exhaustion event.  When this event occurs, the agent will print out a histogram of the first 100 largest types by total number of bytes.
-
-```plain
-Resource exhaustion event: the JVM was unable to allocate memory from the heap.
-ResourceExhausted! (1/0)
-| Instance Count | Total Bytes | Class Name                                    |
-| 18273          | 313157136   | [B                                            |
-| 47806          | 7648568     | [C                                            |
-| 14635          | 1287880     | Ljava/lang/reflect/Method;                    |
-| 46590          | 1118160     | Ljava/lang/String;                            |
-| 8413           | 938504      | Ljava/lang/Class;                             |
-| 28573          | 914336      | Ljava/util/concurrent/ConcurrentHashMap$Node; |
-```
-
-It will also print out a summary of all of the memory spaces in the JVM.
-
-```plain
-Memory usage:
-   Heap memory: init 65011712, used 332392888, committed 351797248, max 351797248
-   Non-heap memory: init 2555904, used 63098592, committed 64815104, max 377790464
-Memory pool usage:
-   Code Cache: init 2555904, used 14702208, committed 15007744, max 251658240
-   PS Eden Space: init 16252928, used 84934656, committed 84934656, max 84934656
-   PS Survivor Space: init 2621440, used 0, committed 19398656, max 19398656
-   Compressed Class Space: init 0, used 5249512, committed 5505024, max 19214336
-   Metaspace: init 0, used 43150616, committed 44302336, max 106917888
-   PS Old Gen: init 43515904, used 247459792, committed 247463936, max 247463936
-```
+Azul Platform Prime JRE does not use the jvmkill agent instead by default uses the -XX:ExitOnOutOfMemoryError flag which terminates the JVM process when an out-of-memory error occurs.
 
 If a [Volume Service][] with the string `heap-dump` in its name or tag is bound to the application, terminal heap dumps will be written with the pattern `<CONTAINER_DIR>/<SPACE_NAME>-<SPACE_ID[0,8]>/<APPLICATION_NAME>-<APPLICATION_ID[0,8]>/<INSTANCE_INDEX>-<TIMESTAMP>-<INSTANCE_ID[0,8]>.hprof`
 
--- a/docs/jre-zing_jre.md
+++ b/docs/jre-zing_jre.md
@@ -1,0 +1,172 @@
+# Azul Platform Prime JRE
+Azul Platform Prime JRE provides Java runtimes developed by Azul team.  Unless otherwise configured, the version of Java that will be used is specified in [`config/zing_jre.yml`][].
+
+<table>
+  <tr>
+    <td><strong>Detection Criterion</strong></td>
+    <td>Unconditional.  Existence of a single bound Volume Service will result in Terminal heap dumps being written.
+      <ul>
+        <li>Existence of a Volume Service service is defined as the <a href="http://docs.cloudfoundry.org/devguide/deploy-apps/environment-variable.html#VCAP-SERVICES"><code>VCAP_SERVICES</code></a> payload containing a service who's name, label or tag has <code>heap-dump</code> as a substring.</li>
+      </ul>
+    </td>
+  </tr>
+  <tr>
+    <td><strong>Tags</strong></td>
+    <td><tt>open-jdk-like-jre=&lang;version&rang;, open-jdk-like-memory-calculator=&lang;version&rang;, jvmkill=&lang;version&rang;</tt></td>
+  </tr>
+</table>
+Tags are printed to standard output by the buildpack detect script.
+
+
+## Configuration
+For general information on configuring the buildpack, including how to specify configuration values through environment variables, refer to [Configuration and Extension][].
+
+The JRE can be configured by modifying the [`config/zing_jre.yml`][] file in the buildpack fork.  The JRE uses the [`Repository` utility support][repositories] and so, it supports the [version syntax][]  defined there.
+
+To use Azul Platform Prime JRE instead of OpenJDK without forking java-buildpack, set environment variable and restage:
+
+```bash
+cf set-env <app_name> JBP_CONFIG_COMPONENTS '{jres: ["JavaBuildpack::Jre::ZingJRE"]}'
+cf restage <app_name>
+```
+
+| Name | Description
+| ---- | -----------
+| `jre.repository_root` | The URL of the Azul Platform Prime repository index ([details][repositories]).
+| `jre.version` | The version of Java runtime to use. Note: version 1.8.0 and higher require the `memory_sizes` and `memory_heuristics` mappings to specify `metaspace` rather than `permgen`.
+| `jvmkill.repository_root` | The URL of the `jvmkill` repository index ([details][repositories]).
+| `jvmkill.version` | The version of `jvmkill` to use.  Candidate versions can be found in the listings for [bionic][jvmkill-bionic].
+| `memory_calculator` | Memory calculator defaults, described below under "Memory".
+
+### Additional Resources
+The JRE can also be configured by overlaying a set of resources on the default distribution. To do this, add files to the `resources/zing_jre` directory in the buildpack fork.
+
+#### JCE Unlimited Strength
+To add the JCE Unlimited Strength `local_policy.jar`, add your file to `resources/zing_jre/lib/security/local_policy.jar`.  This file will be overlayed onto the Azul Platform Prime distribution.
+
+#### Custom CA Certificates
+To add custom SSL certificates, add your `cacerts` file to `resources/zing_jre/lib/security/cacerts`.  This file will be overlayed onto the Azul Platform Prime distribution.
+
+### `jvmkill`
+The `jvmkill` agent runs when an application has experience a resource exhaustion event.  When this event occurs, the agent will print out a histogram of the first 100 largest types by total number of bytes.
+
+```plain
+Resource exhaustion event: the JVM was unable to allocate memory from the heap.
+ResourceExhausted! (1/0)
+| Instance Count | Total Bytes | Class Name                                    |
+| 18273          | 313157136   | [B                                            |
+| 47806          | 7648568     | [C                                            |
+| 14635          | 1287880     | Ljava/lang/reflect/Method;                    |
+| 46590          | 1118160     | Ljava/lang/String;                            |
+| 8413           | 938504      | Ljava/lang/Class;                             |
+| 28573          | 914336      | Ljava/util/concurrent/ConcurrentHashMap$Node; |
+```
+
+It will also print out a summary of all of the memory spaces in the JVM.
+
+```plain
+Memory usage:
+   Heap memory: init 65011712, used 332392888, committed 351797248, max 351797248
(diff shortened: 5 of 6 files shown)


dmikusa-pivotal:
Excellent! Thanks for the submission. At first glance, it looks good. Two points of feedback.

1. We cannot accept unless the CLA is signed. You'll need to click the link where it says "Please click here to be authorized" & follow the instructions. It should update when you've got it all correct.

2. Can you provide some validation instructions? Presumably, I'll need to swap in this JRE in `config/components.yml`, but is there anything else? Are there public downloads of the JRE available? Or do I need to host those on my own network? 

Thanks

ushaazul:
Hi Daniel,
1. Waiting on approval within Azul regarding CLA signing.
2. You will need to swap the JRE in config/components.yml and then push the app
cf push <app_name>  -t 180 -p <jar> -b <githubpath> -m 6000MB  --no-route --health-check-type=process

dmikusa-pivotal:
I totally understand about the CLA. That takes time with companies. When it goes green, I will review & we'll get this merged. Thanks again for the contribution!

ushaazul:
Hi Daniel, 
Our legal team submitted the clearance. I also made some changes after the first pull request. Hope you are able to see both the changes and will accept the merge request

Thanks

ushaazul:
Can we get an ETA on when this change will be merged?

Thanks

dmikusa-pivotal:
It may be a couple of weeks still. There are a number of issues we are trying to address in the next release. I am going to try and squeeze this into that release, but my plan is to review & merge this after the other items are done.

ushaazul:
Thank you so much. Really appreciate the effort.

dmikusa-pivotal:
OK, so I got some time to test this. Some notes.

1. This link didn't work for me. I can view the index page, but the links from the index page are 404's, including the index.yml. 
 
>The public downloads are present at https://cdn.azul.com/zing/bin for cloudfoundry

I don't think this matters though. If I'm understanding correctly, the intent is to have this JDK support disabled by default since this software requires licensing. That means users are to download the software directly from your site, host a local buildpack repo with the software and then use that. Is that correct?

2. I was able to download trial versions of Azul and make a local repo for testing. That went fine. I was able to package up the buildpack and build a couple of apps using it. The buildpack runs and outputs:

>   -----> Downloading Zing JRE 1.8.0_332 from http://localhost:8000/zing22.06.0.0-3-jdk8.0.332-linux_x64.tar.gz (found in cache)
>   Expanding Zing JRE to .java-buildpack/zing_jre (4.5s)

The basics are there and seem to be working.

3. The buildpack installs a tool called the JVM kill agent by default. This does a couple of things, but primarily it watches for OOME's and kills the Java process if one occurs. This tends to work better on CF because the platform can very quickly restart the app to a known good state, rather than trying to struggle through and recover from the OOME which may not ultimately be possible.

https://github.com/cloudfoundry/java-buildpack/blob/main/docs/jre-open_jdk_jre.md#jvmkill

I don't think this is working correctly with the Zing JRE. It detects the OOME when it occurs, but it hangs in the process of trying to dump some final memory statistics about the process (useful for debugging). It seems to be hanging somewhere around [here](https://github.com/cloudfoundry/jvmkill/blob/v1.16.0.RELEASE/jvmkill/src/agentcontroller/poolstats.rs#L65-L66) based on the logging output at the time it hangs.

I'm not sure there's anything you can do with the JVM kill agent, but I don't know off hand why it's failing either, so it's hard to say. What I would probably suggest though is to just not include the JVM kill agent. That is happening automatically because you are basing your ZingJRE implementation off of OpenJDKLike, which [installs the kill agent](https://github.com/cloudfoundry/java-buildpack/blob/main/lib/java_buildpack/jre/open_jdk_like.rb#L41). You could simply remove that line and it'll skip installing the kill agent. You might instead just add some JVM flags to tell the JVM to exit on OOME instead, you can do that with [this](https://github.com/cloudfoundry/java-buildpack/blob/main/lib/java_buildpack/component/java_opts.rb#L102) or [that](https://github.com/cloudfoundry/java-buildpack/blob/main/lib/java_buildpack/component/java_opts.rb#L110).

Another thing installed by the OpenJDKLike implementation by default is the [memory calculator](https://github.com/cloudfoundry/java-buildpack/blob/main/lib/java_buildpack/jre/open_jdk_like.rb#L44). If that makes sense then you're all set. If it does not make sense for this JVM implementation, then you may want to consider removing that and adding JVM flags to tune the JVM's memory settings as well.

You do need to tune JVM memory settings with care because...

1. The buildpack only runs during staging, but the memory limit for an app can change with a simple restart. You may have to use some [bash magic to make the calculation happen at runtime](https://github.com/cloudfoundry/java-buildpack/blob/main/lib/java_buildpack/jre/open_jdk_like_memory_calculator.rb#L55-L59).
2. If an application exceeds its memory limit by even a few bytes, the OOM killer will end the process. It's very unforgiving so it's very important an app does not exceed the memory limit or the app can crash unexpectedly.

I'm not sure how this particular JVM operates so I can't really tell you what's best for some of these design decisions, however, happy to provide more context or answer any additional questions you might have. Just let me know.

ushaazul:
Hi Daniel,
Thanks for testing it.

1. Yes, We needed to change that due to legal issues. Users will now need to download the software from appropriate Azul download channel, host their own repo and use that.
2.  That sounds great
3.  Will need to think about of how best to handle this and will get back to you if I have any questions

ushaazul:
I tried to debug the hang that you were talking about in the earlier comment.
With a simple java pgrm, I did not see any errors. Tried to call all the methods similar to the ones that are being called in impl super::Action for PoolStats

```
# cat MXBeanMemory.java
import java.lang.management.ManagementFactory;
import java.lang.management.MemoryMXBean;
import java.lang.management.MemoryPoolMXBean;
import java.lang.management.MemoryUsage;
import java.util.List;
import java.util.concurrent.TimeUnit;

public class MXBeanMemory{
    private static final int KB = 1024;
    private static final int MB = KB * 1024;
    private static final int GB = MB * 1024;

    public static void main(String[] args) throws InterruptedException {

        for (int i = 0; i < 3; i++) {
            MemoryMXBean memBean = ManagementFactory.getMemoryMXBean() ;
            MemoryUsage heap = memBean.getHeapMemoryUsage();
            System.err.println(String.format("Heap: Init: %d, Used: %d, Committed: %d, Max: %d", heap.getInit() / MB,
                    heap.getUsed() / MB, heap.getCommitted() / MB, heap.getMax() / MB));
            TimeUnit.SECONDS.sleep(1);
        }

        /* For zulujdk8
            Code Cache
            Metaspace
            Compressed Class Space
            PS Eden Space
            PS Survivor Space
            PS Old Gen


            For zulu11
                CodeHeap 'non-nmethods'
                Metaspace
                CodeHeap 'profiled nmethods'
                Compressed Class Space
                G1 Eden Space
                G1 Old Gen
                G1 Survivor Space
                CodeHeap 'non-profiled nmethods'


            For zing 8 and 11
                CodeCache
                GenPauseless New Gen
                GenPauseless Old Gen
                GenPauseless Perm Gen
                Native Heap
         */

        for (int j = 0; j < 3; j++) {
            System.out.println("\n--------------> ");
            List<MemoryPoolMXBean> memoryPoolMXBeanList = ManagementFactory.getMemoryPoolMXBeans();
            System.out.println("Total number of memory pools : " + memoryPoolMXBeanList.size());
            for (int i = 0; i < memoryPoolMXBeanList.size(); i++) {
                MemoryPoolMXBean memoryPoolMXBean = memoryPoolMXBeanList.get(i);
                    System.out.print(memoryPoolMXBean.getName());
                    MemoryUsage usage = memoryPoolMXBean.getUsage();
                    System.err.println(String.format(" Heap: Init: %d, Used: %d, Committed: %d, Max: %d", usage.getInit() / MB,
                            usage.getUsed() / MB, usage.getCommitted() / MB, usage.getMax() / MB));
                    TimeUnit.SECONDS.sleep(1);
            }
        }
    }
```
}

Tried it with Zing 22.06 jdk8 and jdk11.

However when I tried to use rust/cargo. I see the following error
```
error[E0599]: no method named `call_static_object_method` found for struct `JNIEnv` in the current scope
 --> src/lib.rs:9:33
  |
9 |     let memory_mxbean = jni_env.call_static_object_method("java/lang/management/ManagementFactory", get_memory_mxbean_method_id);
  |                                 ^^^^^^^^^^^^^^^^^^^^^^^^^ help: there is an associated function with a similar name: `call_static_method`

For more information about this error, try `rustc --explain E0599`.
```

My rust file looks like this -
```
cat src/lib.rs
use jni::JNIEnv;
use jni::sys::jint;
use jni::objects::JObject;

#[no_mangle]
pub extern "system" fn Java_ch_frankel_blog_rust_Main_doubleRust(jni_env: JNIEnv, _obj: JObject, x: jint) -> jint {
    let mf_class = jni_env.find_class("java/lang/management/ManagementFactory");
    let get_memory_mxbean_method_id = jni_env.get_static_method_id("java/lang/management/ManagementFactory", "getMemoryMXBean", "()Ljava/lang/management/MemoryMXBean;");
    let memory_mxbean = jni_env.call_static_object_method("java/lang/management/ManagementFactory", get_memory_mxbean_method_id);
    x * 2
}

#[no_mangle]
pub extern "system" fn Java_ch_frankel_blog_rust_Main_timesRust(env: JNIEnv, obj: JObject, x: jint) -> jint {
    let state = env.get_field(obj, "state", "I");
    state.unwrap().i().unwrap() * x
}
```
To ensure that we get this get this included in the next release
we could skip installing the killing agent.

dmikusa-pivotal:
>To ensure that we get this get this included in the next release we could skip installing the kill agent.

I'm fine with that. See my previous note, there's info about how you'd go about skipping it.

ushaazul:
Hi Daniel, 
I built the jvmkill agent to see if I could reproduce the hang that you were talking about in an earlier comment

jvmkill agent sources from 
https://github.com/cloudfoundry/jvmkill

I was able to printMemoryUsage stats with no issues. Tried it with Azul's Prime jdk8, 11, 15 and 17.  Pls let me know in case if you can provide any additional details

# /home/ubuntu/zingjdk8/zing22.06.0.0-3-jdk8.0.332-linux_x64/bin/java -agentpath:/home/ubuntu/jvmkill/jvmkill2/jvmkill/target/release/libjvmkill.so=printMemoryUsage=1 GenerateOOM

Resource Exhausted! (1/0)

>>> Memory Pools
Memory usage:
   Heap memory: init 536870912, used 987758592, committed 4196401152, max 4196401152
      Hint: Heap memory is over 95% full. To increase it, increase the container size.
   Non-heap memory: init 183840768, used 10514530304, committed 10514530304, max 16791695360

Memory pool usage:
   CodeCache: init 16777216, used 432640, committed 16777216, max 421527552
   GenPauseless New Gen: init 2097152, used 2097152, committed 2097152, max 4196401152
   GenPauseless Old Gen: init 0, used 811597824, committed 811597824, max 4196401152
   GenPauseless Perm Gen: init 2097152, used 8388608, committed 8388608, max 4196401152
   Native Heap: init 183840768, used 10514530304, committed 10514530304, max 16791695360

>>> Thread Dump
Full thread dump Zing 64-Bit Tiered VM (1.8.0_332-zing_22.06.0.0-b3-product-linux-X86_64 mixed mode [Thu Jul 14 15:19:56 UTC 2022]):

"CompileBrokerThread0" #643 daemon prio=5 os_prio=0 tid=0x620002800000 nid=0x53a3  [ JVM thread_state=_thread_blocked, locked by VM (w/poll advisory bit) waiting for incoming compile requests, polling bits: safep gpgc_clean_new ]
   java.lang.Thread.State: RUNNABLE

"FalconCompThr1" #0 daemon prio=5 os_prio=0 tid=0x620002600000 nid=0x53a2  [ JVM thread_state=_thread_in_vm, locked by VM (w/poll advisory bit) waiting on VM lock 'CompileTask2_lock', polling bits: safep gpgc_clean_new ]
   java.lang.Thread.State: RUNNABLE

"FalconCompThr0" #0 daemon prio=5 os_prio=0 tid=0x620002400000 nid=0x53a1  [ JVM thread_state=_thread_in_vm, locked by VM (w/poll advisory bit) waiting on VM lock 'CompileTask2_lock', polling bits: safep gpgc_clean_new ]
   java.lang.Thread.State: RUNNABLE

"C1CompThr1" #0 daemon prio=5 os_prio=0 tid=0x620002200000 nid=0x53a0  [ JVM thread_state=_thread_in_vm, locked by VM (w/poll advisory bit) waiting on VM lock 'CompileTask1_lock', polling bits: safep gpgc_clean_new ]
   java.lang.Thread.State: RUNNABLE

"C1CompThr0" #0 daemon prio=5 os_prio=0 tid=0x620002000000 nid=0x539f  [ JVM thread_state=_thread_in_vm, locked by VM (w/poll advisory bit) waiting on VM lock 'CompileTask1_lock', polling bits: safep gpgc_clean_new ]
   java.lang.Thread.State: RUNNABLE

"Service Thread" #483 daemon prio=5 os_prio=0 tid=0x620001e00000 nid=0x539e  [ JVM thread_state=_thread_blocked, locked by VM (w/poll advisory bit) waiting on VM lock 'Service_lock', polling bits: safep gpgc_clean_new ]
   java.lang.Thread.State: RUNNABLE

"Signal Dispatcher" #451 daemon prio=5 os_prio=0 tid=0x620001c00000 nid=0x539d  [ JVM thread_state=_thread_in_vm, locked by VM (w/poll advisory bit) waiting on VM lock 'VMOperationRequest_lock', polling bits: safep ]
   java.lang.Thread.State: RUNNABLE

"Object Monitor Thread" #419 daemon prio=5 os_prio=0 tid=0x620001a00000 nid=0x539c  [ JVM thread_state=_thread_blocked, locked by VM (w/poll advisory bit) sleeping, polling bits: safep gpgc_clean_new ]
   java.lang.Thread.State: RUNNABLE

"Surrogate Locker Thread" #387 daemon prio=5 os_prio=0 tid=0x620001800000 nid=0x539b  [ JVM thread_state=_thread_in_vm, locked by VM (w/poll advisory bit) waiting on VM lock 'SLT_lock', polling bits: safep ]
   java.lang.Thread.State: RUNNABLE

"Finalizer" #355 daemon prio=5 os_prio=0 tid=0x620001600000 nid=0x539a  [ JVM thread_state=_thread_in_vm, locked by VM (w/poll advisory bit) waiting on VM lock 'java.lang.ref.ReferenceQueue$Lock', polling bits: safep gpgc_clean_new ]
   java.lang.Thread.State: WAITING (on object monitor)
	at java.lang.Object.wait(Native Method)
	- waiting on <0x0000020000c11810> (a java.lang.ref.ReferenceQueue$Lock)
	at java.lang.ref.ReferenceQueue.remove(ReferenceQueue.java:144)
	- locked <0x0000020000c11810> (a java.lang.ref.ReferenceQueue$Lock)
	at java.lang.ref.ReferenceQueue.remove(ReferenceQueue.java:165)
	at java.lang.ref.Finalizer$FinalizerThread.run(Finalizer.java:216)

"Reference Handler" #323 daemon prio=5 os_prio=0 tid=0x620001400000 nid=0x5399  [ JVM thread_state=_thread_in_vm, locked by VM (w/poll advisory bit) waiting on VM lock 'java.lang.ref.Reference$Lock', polling bits: safep gpgc_clean_new ]
   java.lang.Thread.State: WAITING (on object monitor)
	at java.lang.Object.wait(Native Method)
	- waiting on <0x0000020000c10e70> (a java.lang.ref.Reference$Lock)
	at java.lang.Object.wait(Object.java:502)
	at java.lang.ref.Reference.processThroughPending(Reference.java:281)
	- locked <0x0000020000c10e70> (a java.lang.ref.Reference$Lock)
	at java.lang.ref.Reference.tryHandlePending(Reference.java:204)
	at java.lang.ref.Reference$ReferenceHandler.run(Reference.java:177)

"main" #5 prio=5 os_prio=0 tid=0x620000000000 nid=0x538f  [ JVM thread_state=_thread_in_native, locked by VM (w/poll advisory bit) running, polling bits: safep ]
   java.lang.Thread.State: RUNNABLE
	at GenerateOOM.generateOOM(GenerateOOM.java:20)
	at GenerateOOM.main(GenerateOOM.java:11)

"VM Thread" os_prio=0 tid=0x620001200000 nid=0x5398

"GC task thr#0 (GenPauselessGC)" os_prio=0 tid=0x620000400000 nid=0x5391

"GC task thr#1 (GenPauselessGC)" os_prio=0 tid=0x620000600000 nid=0x5392

"GC task thr#0 (GenPauselessGC)" os_prio=0 tid=0x620000800000 nid=0x5393

"VM Periodic Task Thread" os_prio=0 tid=0x620002a00000 nid=0x53a4

JNI global references: 1123

Heap
GenPauselessHeap
Max 4002M, Capacity 4002M, Used 944M
Virtual Max: New 4002M, Old/Perm 4002M
NewCollector thread: 7 GC cycle: 17-12 state: ConcurrentMarking [0x5]
OldCollector thread: 8 GC cycle: 16-5 state: ConcurrentRefProcessing [0x7]


jvmkill is killing current process
Killed

lib/java_buildpack/jre/open_jdk_like.rb
```diff
@@ -38,10 +38,8 @@ def command
       # (see JavaBuildpack::Component::ModularComponent#sub_components)
       def sub_components(context)
         [
-          JvmkillAgent.new(sub_configuration_context(context, 'jvmkill_agent')),
```
dmikusa-pivotal, line 41: We can't remove this here as it'll break other things. Your class ZingJRE inherits from this class though, so what you need to do is to override this method, `sub_components` in your class and supply a list that does not include JVM Kill agent.

dmikusa-pivotal:
As far as the killagent goes, that repo is in a weird state. A past developer on the project did a major rewrite on the main branch but that never actually got released. He's no longer working on the project and I don't know the state of that work. 

If you want to recompile, you need to checkout the latest tagged release & compile from that. That in itself has problems because some of the dependencies are very old, so it doesn't build as is with recent Rust versions. I'm working on updating that and getting something we can build again. I just haven't had time to finish that work yet.

Just doing a second pass on testing with the kill agent. I'm using zing22.06.0.0-3-jdk8.0.332-linux_x64.tar.gz which I got from your trial downloads. It installs OK & and the app starts OK. I have a demo app that just adds new `int[]` arrays until it exhausts the heap. This triggers the OOME as expected and the kill agent detects it, but that's where I was getting the hang. Normally the kill agent will kill the process which causes the CF app to be automatically restarted. In this case, it just hangs and never ends up killing the process, thus the app never restarts (the app is also no longer in a place where it responds to traffic).

```
  2022-07-19T15:05:15.91-0400 [APP/PROC/WEB/0] OUT 2022-07-19 19:05:15.917  INFO 36 --- [io-8080-exec-10] c.v.m.s.jmw.MemoryWasterAPIController    : Going to create 50000 junk of int[1000]...
   2022-07-19T15:05:17.35-0400 [APP/PROC/WEB/0] ERR Resource exhaustion event: the JVM was unable to allocate memory from the heap.
   2022-07-19T15:05:17.35-0400 [APP/PROC/WEB/0] ERR ResourceExhausted! (1/0)
   2022-07-19T15:05:18.95-0400 [APP/PROC/WEB/0] OUT | Instance Count | Total Bytes | Class Name                                                                                                       |
   2022-07-19T15:05:18.95-0400 [APP/PROC/WEB/0] OUT | 141124         | 559637544   | [I                                                                                                               |
   2022-07-19T15:05:18.95-0400 [APP/PROC/WEB/0] OUT | 7139           | 6601568     | Ljava/lang/Class;                                                                                                ...
   2022-07-19T15:05:19.08-0400 [APP/PROC/WEB/0] OUT | 119            | 6664        | Lorg/thymeleaf/engine/HTMLElementName;                                                                           |
   2022-07-19T15:05:19.08-0400 [APP/PROC/WEB/0] OUT | 166            | 6640        | Lorg/springframework/http/MediaType;                                                                             |
   2022-07-19T15:05:19.08-0400 [APP/PROC/WEB/0] OUT | 166            | 6640        | Lorg/springframework/core/type/classreading/SimpleMethodMetadataReadingVisitor$Source;                           |
   2022-07-19T15:05:19.11-0400 [APP/PROC/WEB/0] OUT Memory usage:  <------- ## Hung Here ##
```

For what it's worth, I'm getting this error also when the app starts up. It seems possibly related.

```
   2022-07-19T15:00:22.13-0400 [APP/PROC/WEB/0] ERR java.lang.NullPointerException
   2022-07-19T15:00:22.13-0400 [APP/PROC/WEB/0] ERR at io.micrometer.core.instrument.binder.jvm.JvmGcMetrics.countPoolSizeDelta(JvmGcMetrics.java:199)
   2022-07-19T15:00:22.13-0400 [APP/PROC/WEB/0] ERR at io.micrometer.core.instrument.binder.jvm.JvmGcMetrics.lambda$bindTo$4(JvmGcMetrics.java:159)
   2022-07-19T15:00:22.13-0400 [APP/PROC/WEB/0] ERR at sun.management.NotificationEmitterSupport.sendNotification(NotificationEmitterSupport.java:156)
   2022-07-19T15:00:22.13-0400 [APP/PROC/WEB/0] ERR at sun.management.GarbageCollectorImpl.createGCNotification(GarbageCollectorImpl.java:143)
   2022-07-19T15:00:22.13-0400 [APP/PROC/WEB/0] ERR Exception in thread "Service Thread" java.lang.AssertionError: Error in invoking listener
   2022-07-19T15:00:22.13-0400 [APP/PROC/WEB/0] ERR at sun.management.NotificationEmitterSupport.sendNotification(NotificationEmitterSupport.java:159)
   2022-07-19T15:00:22.13-0400 [APP/PROC/WEB/0] ERR at sun.management.GarbageCollectorImpl.createGCNotification(GarbageCollectorImpl.java:143)
```

I'm certainly not trying to cast any blame here. It's very likely that there is something with the killagent, not handling a return value correct or handling the state of things properly, or possibly something specific to the environment on CF where the app runs.

Given that you can support most of the functionality with JVM flags like `-XX:+ExitOnOutOfMemoryError` and  `-XX:+HeapDumpOnOutOfMemoryError`/`-XX:HeapDumpPath=`, it may not be worth the effort to debug this failure and instead go with these JVM flags instead, documenting the difference. You can add these flags in your subclass of OpenJDKLike, just like you can remove the killagent there.

Things that are remaining:

1. Move the removal of the kill agent into your subclass
2. Add JVM flags by default such that it exits on OOME. Optionally, add heap dump on OOME, but that's more complicated so I'd be OK not including that.
3. Update the documentation to include these differences in behavior.

ushaazul:
Thanks for testing it and your explanation.
Coming back to the remaining changes. I am a little confused. Could you pls clarify and let me know if it doesn't look right.
I intend to make the following changes to  in response to items listed above
For item 1 - lib//java_buildpack/jre/zing_jre.rb
```
class ZingJRE < OpenJDKLike
        
        def sub_components(context)
          [
            OpenJDKLikeJre.new(sub_configuration_context(context, 'jre')
                               .merge(component_name: self.class.to_s.space_case)),
            OpenJDKLikeSecurityProviders.new(context)
          ]
        end 
```
item 1 second changes - lib//java_buildpack/jre/open_jdk_like.rb
```
def sub_components(context)
        [
          JvmkillAgent.new(sub_configuration_context(context, 'jvmkill_agent')),
          OpenJDKLikeJre.new(sub_configuration_context(context, 'jre')
                               .merge(component_name: self.class.to_s.space_case)),
          OpenJDKLikeMemoryCalculator.new(sub_configuration_context(context, 'memory_calculator')),
          OpenJDKLikeSecurityProviders.new(context)
        ]
      end
```
For item2 -  spec/java_buildpack/framework/java_opts_spec.rb
```110   it 'does not include $JAVA_OPTS without from_environment' do
    111     component.release
    112     expect(java_opts).not_to include('$JAVA_OPTS')
    113     expect(java_opts).to include('-XX:OnOutOfMemoryError=kill\ -9\ \%p')```

**Comment by dmikusa-pivotal** [comment]
For item 1 - Yes, that looks right.

For item 2 - You need to override `release` in your class, something like this

```
      def release
        super
        @droplet.add_preformatted_options "-XX:+ExitOnOutOfMemoryError"
      end
```

Using `'-XX:OnOutOfMemoryError'` isn't recommended, not for killing the process anyway. It can cause problems when there's low/no memory available in a container cause it can't spawn the kill process. The option in my example is newer but available in Java 1.8+ (recent Java 1.8's anyway) and is meant for specifically this situation.

Again, you could try to add in heap dump options but that's probably better to do in a follow-up feature if someone requires it. To do that you would need to check for a bound volume service & set it up to write to that location. This is because you have to write the heap dump somewhere durable or it'll disappear when the app exits (file system in CF is ephemeral).

ushaazul:
I have submitted changes as per your suggestion. Pls let me know if we need anything else.

dmikusa-pivotal:
I did some more testing and had a few minor things come up. I had to adjust how the command was being overridden, cause the way it was coded had resulted in the arguments being duplicated. I'm going to merge your PR and follow after that with a small PR to polish things up.

As far as timing, I should be able to get this out very soon. We do have to cut a patch release first with updates for the July 2022 quarterly updates. You'll see that here any time now, and then I'll merge this PR and #957, then I'll cut a new minor release with this functionality. That might be Friday, but more likely will be Monday.

Thanks again for the PR!

ushaazul:
Thanks for the update and the ETA for the patch.
Can we get more details on the release/patch number once it is available

dmikusa-pivotal:
Thanks again for the PR! Final update on timing. We have two more PRs to merge & are targeting the release on Friday.

PR #956, opened 2022-07-07
Remove Takipi Agent

dmikusa-pivotal:
The download site for the Takipi Agent has gone dark, and attempts to alert the vendor and rectify the situation have failed. This component will no longer be enabled with the default distribution of the Java buildpack. You can re-enable it if you are able to host your own repository for the downloads.

If/when the vendor is able to resolve the situation, we can re-enable the agent by default in the buildpack.

### DIFF
--- a/config/components.yml
+++ b/config/components.yml
@@ -73,6 +73,5 @@ frameworks:
   - "JavaBuildpack::Framework::SpringInsight"
   - "JavaBuildpack::Framework::SkyWalkingAgent"
   - "JavaBuildpack::Framework::YourKitProfiler"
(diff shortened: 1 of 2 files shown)


M-Tsur:
@dmikusa Hi, I'm am part of OverOps (f.k.a Takipi). We have addressed the issue as it was a permissions problem on our end due to some organization changes. How can we get this enabled again?
If needed, we can open a PR to revert the change.

dmikusa:
@M-Tsur - thanks for reaching out. Yes, if it's working again, please open a PR to revert this change. Thanks

Issue #892, opened 2021-07-14
Pinned Offline Buildpack chooses non included versions


benjaminguttmann-avtq:
Hi there, 

I am currently seeing a strange behavior when using offline buildpacks, where the version used is not included in the offline package and therefore fails to stage the app. 

Steps to reproduce the issue:

```
git clone https://github.com/cloudfoundry/java-buildpack.git
cd java-buildpack
git checkout v4.37
bundle exec rake clean package OFFLINE=true PINNED=true STACK=cflinuxfs3 VERSION="v4.37"
mv build/java-buildpack-offline-v4.37.zip ./java-buildpack-offline-v4.37.zip
git checkout v4.39
bundle exec rake clean package OFFLINE=true PINNED=true STACK=cflinuxfs3 VERSION="v4.39"
cf login
cf create-buildpack java-offline-buildpack-v4-39 ./build/java-buildpack-offline-v4.39.zip 1
cf create-buildpack java-offline-buildpack-v4-37 ./build/java-buildpack-offline-v4.37.zip 1
cf push spring-music -f manifest.yml -b java-offline-buildpack-v4-37
```

Manifest looks like this:

```
---
applications:
- name: spring-music
  instances: 1
  memory: 1G
  env:
    JBP_CONFIG_OPEN_JDK_JRE: '{ jre: { version: 1.8.+}}'
```

We are using [spring-music](https://github.com/cloudfoundry-samples/spring-music) to test the behavior. 

In our env we can see that the staging fails due to this error:

```
   2021-07-14T08:27:00.95+0000 [STG/0] ERR [Buildpack]                      ERROR Finalize failed with exception #<RuntimeError: Unable to find cached file for https://java-buildpack.cloudfoundry.org/openjdk/bionic/x86_64/bellsoft-jre8u292%2B10-linux-amd64.tar.gz>
   2021-07-14T08:27:00.95+0000 [STG/0] ERR Unable to find cached file for https://java-buildpack.cloudfoundry.org/openjdk/bionic/x86_64/bellsoft-jre8u292%2B10-linux-amd64.tar.gz
   2021-07-14T08:27:00.96+0000 [STG/0] ERR Failed to compile droplet: Failed to run finalize script: exit status 1
```

We would expect the buildpack to choose the included version `1.8.0_282`.

Beside that, for me `v4.37` pins version to `11.0.11_9` but according to the release notes it should be `11.0.10_9`.

dmikusa-pivotal:
[...]

Issue #886, opened 2021-06-15
Make JDK 11 as Default and Deprecate JDK 8


ywei2017:
We have been trying to promote JDK 11, instead of the default JDK 8. But most dev teams don't do anything, so they use JDK 8 as default.

I was wondering of the possibility of moving the default to JDK 11, since time has progressed. Would that be a good idea, or there are many reasons against it?

We really would like to retire JDK 8 all together. What is the time horizon for that?

Thanks
Yansheng

dmikusa-pivotal:
It's not something that has been discussed yet, however, I do appreciate feedback from you and other users on the subject. If you are interested in this change, please 👍  this issue.

In the meantime, if you would like to make it the default you can. You would just need to edit [this line in the config file](https://github.com/cloudfoundry/java-buildpack/blob/main/config/open_jdk_jre.yml#L20) then [follow the instructions to package the buildpack](https://github.com/cloudfoundry/java-buildpack#building-packages). The "online" version would be the same as what you get through the Releases page. You can build an "offline" version if your users are building in environments without Internet access, or if you want to speed up builds by not requiring downloads at the cost of having a larger buildpack, you can also use the "offline" variant.

mayrstefan:
Changing the default raises two questions:
1. When will Bellsoft EOL the different Java versions?
2. What would be the buildpack policy to change the default between LTS versions? Next would be Java 17

Looking at Adoptium we will get
* Java 8 at least until 2026
* Java 11 at least until 2024
* Java 17 no statement yet

So in 2025 we have at least two choices: back to Java 8 because it is still supported or move forward to Java 17. Or Java 23 or whatever will be the next LTS release.

dmikusa-pivotal:
>When will Bellsoft EOL the different Java versions?

https://bell-sw.com/pages/roadmap/  (far right column).

Java 8 is in a weird spot. It's technically supported longer than both Java 11 and 17. It looks like Adoptium is similar. That is definitely a concern with changing the default. That would necessitate another default change to 17 at some point, whereas if we stick with 8 then no changes for even longer.

>What would be the buildpack policy to change the default between LTS versions? Next would be Java 17

I think this is an open question because in the history of the buildpack we haven't had to do that yet. Regardless of the default value, we'll continue supporting Java 8, 11, and 17 through their lifecycle, so even if the default changes to Java 11 you'll still be able to use Java 8.

[...]

PR #957, opened 2022-07-15
Add default functionality & refactor packaging

### LINKED ISSUE #938: Allow offline builds to choose included java versions
The offline buildpack is very hefty because it includes three JDKs. It would be nice if instead of hard coding the three version support users were able to specify which versions they wanted instead. I suggest having a variable that defaults to the three versions but can be changed by the user.

dmikusa-pivotal:
1. Adds an environment variable that can be used by operations teams to set default values. Operator defaults override buildpack defaults, but are overridden by application developer values. Operator defaults are set by exposing an environment variable of the pattern `JBP_DEFAULT_<config-file>='<inline-yaml>'`. So to set a default Java version, `JBP_DEFAULT_OPEN_JDK_JRE='{ jre: { version: 11.+ }}'`.

2. Adds `config/packaging.yml`. This contains the list of component data. Each component gets a name, CVE notes link and release notes link. This information is used by the `rake versions` command to generate a nice set of release notes for each package. This information was refactored out of the scripts (was previously hard-coded).

3. Each `config/<jre>.yml` file has been given a `jre.version_lines` property. This is a list of the supported version lines for that JRE. It defaults to using pattern matching to pick the latest 1.8, 11 and 17 versions. If you are packaging your own version of the buildpack and want to limit the version lines included, perhaps to reduce the size of the generated buildpack file, you can edit this list and remove entries. Theh package script will only include what is listed in this array.

4. When pinning versions, the packaging script will now update each `config/<jre>yml` file's `jre.version_lines` list with the pinned versions, in addition to `jre.version`. This allows you to have a list of every version line bundled with the buildpack when a release is cut (previously, only 1.8 was committed to source control). This helps when rebuilding a version of the Java buildpack at a later date, as you'll have the exact set of Java versions which should be included.

5. You may change the default version of Java that is selected during a build of the Java buildpack by setting the `config/<jre>.yml` file's `jre.version` value. This needs to match one of the entries in `jre.version_lines`. Then follow the standard packaging steps the the buildpack that's generated will use the selected version of Java by default.

Resolves #938 #892 #886 

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,12 @@ $ cf set-staging-environment-variable-group '{"JBP_DEFAULT_OPEN_JDK_JRE":"{jre: 
 $ cf set-staging-environment-variable-group '{"JBP_CONFIG_REPOSITORY": "{default_repository_root: \"http://repo.example.io\" }"}'
 ```
 
+3. To change the default JVM vendor across all applications on a foundation. Be careful to ensure that your JSON is properly escaped.
+
+```bash
+$ cf set-staging-environment-variable-group '{"JBP_CONFIG_COMPONENTS": "{jres: [\"JavaBuildpack::Jre::ZuluJRE\"]}"}'
+```
+
 ### Application Developer
 
 1. To change the default version of Java to 11 and adjust the memory heuristics then apply this environment variable to the application.
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Cloud Foundry Java Buildpack
 
-The `java-buildpack` is a [Cloud Foundry][] buildpack for running JVM-based applications.  It is designed to run many JVM-based applications ([Grails][], [Groovy][], Java Main, [Play Framework][], [Spring Boot][], and Servlet) with no additional configuration, but supports configuration of the standard components, and extension to add custom components.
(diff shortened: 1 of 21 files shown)


README.md
```diff
@@ -21,7 +21,21 @@ The following are _very_ simple examples for deploying the artifact types that w
...
+  - If you are an operator that wishes to override configuration across a foundation, you may do this by setting environment variable group entries that begin with a prefix of `JBP_DEFAULT`.
+  - If you are an application developer that wishes to override configuration for an individual application, you may do this by setting environment variables that begin with a prefix of `JBP_CONFIG`. 
+
+For example, to change the default version of Java to 11 across all applications.
+
+```bash
+$ cf set-staging-environment-variable-group '{"JBP_DEFAULT_OPEN_JDK_JRE":"{jre: {version: 11.+ }}"}'
+```
+
+To change the default version of Java to 11 and adjust the memory heuristics then apply this environment variable to the application.
```
pivotal-david-osullivan, line 38: Minor, for readability - since these 2 examples are of the previous listed 2 levels of overrides, you might highlight that the second is for an 'individual application' to specifically tie it back to override #2. You could even number the overrides & examples.

dmikusa-pivotal: Sounds good. I will try to elaborate more on that in the example. Thanks

dmikusa-pivotal: Done.

PR #972, opened 2022-09-27
Updated Contrast agent version

FrankGasparovic:


dmikusa:
Are you able to sign the CLA? Apologizes, but we have to have every commit pass the EasyCLA check no matter how small.

Aside from that, this would be a major bump for the agent version. We need to be careful doing this as it can impact existing users.

I see the last version of the 3.x branch was published July 6, 2022 which isn't that long ago. It also supports up through Java 17. My inclination would be to let this sit for a bit and see if other users want us to bump the default version as well.

For what it's worth, you can individually change the version to the 4.x branch by setting that with a `JBP_CONFIG` or `JBP_DEFAULT` config setting [see here](https://github.com/cloudfoundry/java-buildpack#configuration-and-extension).

FrankGasparovic:
@dmikusa The CLA has been signed. 

There are a number of new features and bug fixes outlined [here](https://docs.contrastsecurity.com/en/java-agent-release-notes-and-archive.html) which were introduced in the 4.x java agents which is why we recommend all customers to upgrade.

dmikusa (approved):

dmikusa:
Thanks @FrankGasparovic I'm onboard and I've approved this. I'm going to defer to @pivotal-david-osullivan on merging & releasing.

PR #974, opened 2022-10-07
Handle Dynatrace API Token in the sanitizer

rbamberger:
Masking the private part of the API Token

Fix for Issue [#967](https://github.com/cloudfoundry/java-buildpack/issues/967)

### DIFF
--- a/lib/java_buildpack/util/sanitizer.rb
+++ b/lib/java_buildpack/util/sanitizer.rb
@@ -22,11 +22,32 @@ class String
   #
   # @return [String] the sanitized uri
   def sanitize_uri
+    keywords = /key|password|username|cred[entials]*[s]*|password|token|api[-_]token|api|auth[entication]*|access[-_]token|secret[-_]token/i
+
     rich_uri          = URI(self)
     rich_uri.user     = nil
     rich_uri.password = nil
-    rich_uri.query = rich_uri.query&.gsub(/(Api-Token=dt\w*\.\w*)\.\w*/, '\1.REDACTED')
+
+    if(rich_uri.query)
+      params = Hash[URI.decode_www_form rich_uri.query]
+
+      query_params = ""
+
+      params.each do |key,value|
+        match = key.match(keywords)
+
+        if(match)
+          if(match[0] == "Api-Token" && value =~ /dt\w*/)
+            params[key] = value.gsub(/(dt\w*\.\w*)\.\w*/, '\1.REDACTED')
+          else
+            params[key] = "***"
+          end
+        end
+
+        query_params += key + "=" + params[key] + "&"
+      end
+      rich_uri.query = query_params.chop
+    end
     rich_uri.to_s
   end
-
 end
--- a/lib/java_buildpack/util/sanitizer.rb
+++ b/lib/java_buildpack/util/sanitizer.rb
@@ -18,36 +18,55 @@
 # A mixin that adds the ability to turn a +String+ into sanitized uri
 class String
 
+  # Takes the uri query params and strips out credentials
+  #
+  # @return [String] the sanitized query params
+  def handle_params(params)
+    keywords = /key
+               |password
+               |username
+               |cred(ential)*(s)*
+               |password
+               |token
+               |api[-_]token
+               |api
+               |auth(entication)*
+               |access[-_]token
+               |secret[-_]token/ix
+
+    query_params = ''
+
+    params.each do |key, value|
+      match = key.match(keywords)
+
+      if match
+        params[key] = if match[0] == 'Api-Token' && value =~ /dt\w*/
+                        value.gsub(/(dt\w*\.\w*)\.\w*/, '\1.REDACTED')
+                      else
+                        '***'
+                      end
+      end
+
+      query_params += key + '=' + params[key] + '&'
+    end
+
+    query_params
+  end
+
   # Takes a uri and strips out any credentials it may contain.
   #
   # @return [String] the sanitized uri
   def sanitize_uri
-    keywords = /key|password|username|cred[entials]*[s]*|password|token|api[-_]token|api|auth[entication]*|access[-_]token|secret[-_]token/i
-
     rich_uri          = URI(self)
     rich_uri.user     = nil
     rich_uri.password = nil
 
-    if(rich_uri.query)
-      params = Hash[URI.decode_www_form rich_uri.query]
-
-      query_params = ""
-
-      params.each do |key,value|
-        match = key.match(keywords)
-
-        if(match)
-          if(match[0] == "Api-Token" && value =~ /dt\w*/)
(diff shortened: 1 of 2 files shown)


dmikusa (changes requested):
I think this is a good start, but it's really only addressing the specific case of dynatrace, we should probably make this a general solution. I can imagine that there are other things that should be sanitized from download URLs.

Can you do something like this?

1. Make a list of query params to sanitize. The list can be strings or regex patterns.
2. For each item in the list, go through the query params, and if the param key matches (case insensitive), then set the value to `****`. I don't think we need to care about the value. If the query param key matches, then we just overwrite it.
3. Add some unit tests to show this working.

Some keys for the list, just trying to think of possibly sensitive query params:

```
username
password
token
api
key
secret
access[-_]token
secret[-_]token
api[-_]key
api[-_]token
auth[entication]*
cred[s]*
```

and feel free to add anything else that comes to mind.

rbamberger:
I have also already thought of implementing a general solution. However, this is a problem for which we need a solution quickly. This was the least invasive fix from our point of view.

TimGerlach:
To speed things up a little bit I would also appreciate if this could be implemented in two phases so that we get the Dynatrace API token masking with this PR and a more general solution with a follow-up PR.
We do have some time pressure in the context of security requirements.

dmikusa:
My vote would be for a proper fix, I don't think it's a huge request. That would also add in a solution that's actually generic, right now, this PR is modifying core parts of the buildpack with DT-specific logic.

TimGerlach:
Hi @dmikusa,
Is there any chance to get another PR review from you?

lib/java_buildpack/util/sanitizer.rb
```diff
@@ -22,11 +22,32 @@ class String
...
+      params = Hash[URI.decode_www_form rich_uri.query]
+
+      query_params = ""
+
+      params.each do |key,value|
+        match = key.match(keywords)
+
+        if(match)
+          if(match[0] == "Api-Token" && value =~ /dt\w*/)
+            params[key] = value.gsub(/(dt\w*\.\w*)\.\w*/, '\1.REDACTED')
```
dmikusa, line 41: What is the need for this inner `if` block? Why can't we just have everything be masked in the same way?

dmikusa (changes requested):
Apologizes. I missed the updated, thanks for the ping.

This generally looks good, but it needs at least a few unit tests to validate the behavior.

lib/java_buildpack/util/sanitizer.rb
rbamberger: Thanks for your review. The if is needed because we only want to mask parts of the Dynatrace API Token. The Tokens have a public and a private part. Masking only the private part helps in identifying which token was used for staging the app by looking at the logs.

rbamberger: Just one note from my side. My concern is that if we manipulate the URI we might end up with similar Hashes for different files, possibly breaking other stuff in the Buildpack. We are touching the Buildpack Code at a very low level here. This Buildpack is a very complex piece of software and we should be aware of possible side effects.

The sanitize_uri function is used in different places, mainly for Logging URL's and generating a filename for the cached file artifacts. (File, Etag, Last-Modified)

The Download Cache itself is used for a lot of things. Main Use Case is for downloading Stuff directly by using the download function in the Base Component but also in some other components like External Config or the Repository Index. 

But only someone with a lot of knowledge about all the mechanisms in the Buildpack can judge possible side effects. I can't.

The pull request is also missing the tests.

rbamberger: Hi @dmikusa,
I've updated the code to fulfill the RuboCop recommendations and also added a test that includes the tests. Please have another look. Thanks in advance.

meibensteiner:
@dmikusa Is there any new development here? We need to get this merged for a customer asap!

lib/java_buildpack/util/sanitizer.rb
dmikusa: Thanks for this PR & for bearing with me through the process.

I'm not a huge fan of having the special case part in there, but mostly because I think it could be potentially difficult to handle. If there's one special case, then that invites the potential for others. We can cross that bridge if it becomes an issue though.

>The sanitize_uri function is used in different places, mainly for Logging URL's and generating a filename for the cached file artifacts. (File, Etag, Last-Modified)

This is a good point. If it's caching a URL with the masked value, that would be less unique and could introduce collisions (i.e. the masked value changes, but it's not reflected in the URL for caching purposes). 

Since the buildpack cache is scoped to your application, I'm not sure this would be a huge issue. Taking Dynatrace for example. The user would need to perform a build with one key and then switch the key and rebuild. In that case, I think you'd get a collision. It's hard to say if that would be a problem. In most cases where everyone gets the same set of binaries, it wouldn't be. If the binary being fetched was customized and unique to the particular URL then that could be an issue.

I could see this being handled in two ways.

1. We could adjust the caching to pull the real URL, not the masked URL, that way we have the full and unique URL for caching purposes. The trick with this approach is that we'd need to be really careful to not have the actual URL saved somewhere in the cache as that would be another form of information leak.
2. Since most dependencies are not impacted by this, as they are not unique per app/user/token, we could handle this for the particular framework. Essentially, the framework having domain knowledge would know if the dependency might be susceptible to this problem. It could then clear the cache when it detects a change to the secret/token/password/etc.

At any rate, I'm not sure if we need to deal with this now or if we can wait and follow up with another PR. I'd defer to you on that, given you know how this might impact Dynatrace users. What do you think?

dmikusa:
@meibensteiner Hi, posted some thoughts. Sorry for the delay. I don't work full time on this anymore, just spare time/OSS capacity.

@pivotal-david-osullivan - FYI for integration testing/release and also for a second opinion.

lib/java_buildpack/util/sanitizer.rb
rbamberger: @dmikusa 
I agree that the special handling of the DT Token in that place is not a good idea. I suggest removing that part again. Then we have at least the problem with logging secrets solved for now.

rbamberger:
@dmikusa Hi, I've removed the special handling for the Dynatrace token. Please have another look.

TimGerlach:
Dear @dmikusa, @pivotal-david-osullivan,
Can you please consider another review of this Pull Request?
We're currently stuck in security audits and need to provide an ETA for masking the tokens in the staging logs any time soon.

Reviewing the PR should probably not take longer than 10 minutes.

Thanks for your support!

pivotal-david-osullivan:
Yes, we'll take another look at this ASAP, thanks!

pivotal-david-osullivan (approved):

anthonydahanne (approved):

dmikusa (approved):

Issue #980, opened 2022-11-11
Azure Application Insights Agent is outdated


mayrstefan:
The buildpack includes an outdated version of Azure Application Insights Agent which was released over two years ago.
Version 2.x should be replaced by 3.x We can find an upgrade guide in https://learn.microsoft.com/en-us/azure/azure-monitor/app/java-standalone-upgrade-from-2x
I see one breaking change when upgrading the agent. The old version used an instrumentation key which could be set on the command line. The new version changes to a connection string which can be configured as an environment variable or a json file

mayrstefan:
I found https://learn.microsoft.com/en-us/azure/azure-monitor/app/java-standalone-config#connection-string which states that we still can use a system property `applicationinsights.connection.string`. So I prepared a PR to make use of a `connection_string` instead of `instrumentation_key` in a user-provided-service

dmikusa:
I saw your related PRs. Thanks for sending them over.

I'm not super familiar with Azure App Insights. I suspect because it's a major version of the agent that if we change the default, we'll end up with the possibility of breaking users. Or are you saying that with the link you referenced users should not notice a difference after upgrading?

The challenge with the Java buildpack and changing defaults is that we should really consider it a breaking change because it can break user applications. Thus we have to either slowly roll out the change, i.e. put in a warning that the change is coming, wait for a while, then make the change. 

That or discuss possibly cutting a new major version of the buildpack. I suspect if there are enough different dependencies that we need to bump then perhaps a new major version would make sense. We do also have a PR for [Contrast Security](https://github.com/cloudfoundry/java-buildpack/pull/972). We've also been trying to slow deprecate some code and functionality that has gone EOGS in upstream projects, like Redis session store for Tomcat and the Spring Cloud Connectors and Spring Auto Reconfiguration support. Perhaps it's time to start working towards a new major?

At any rate, the other side of the coin here is that we fully support newer versions of dependencies, they are just not the defaults. In addition, you can easily package your own defaults by packaging your own copy of the buildpack and you can override defaults with the [newly added operator config overrides](https://github.com/cloudfoundry/java-buildpack#operator). So if you have an immediate need to see this change, I would suggest looking at one of those options for a short-term fix.

mayrstefan:
[...]

PR #981, opened 2022-11-13
Add connection string for Azure Application Insights Agent 3.x

mayrstefan:
Improvements to Azure Application Insights Agent to make version 3.x usable using a connection string (see also #980). Also added some initial documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,7 @@ The buildpack supports extension through the use of Git repository forking. The 
 * Standard Frameworks
   * [AppDynamics Agent](docs/framework-app_dynamics_agent.md) ([Configuration](docs/framework-app_dynamics_agent.md#configuration))
   * [AspectJ Weaver Agent](docs/framework-aspectj_weaver_agent.md) ([Configuration](docs/framework-aspectj_weaver_agent.md#configuration))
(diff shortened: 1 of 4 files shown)


pivotal-david-osullivan:
Hi @mayrstefan, thanks for this PR, I did some testing of the changes and I cannot get the framework to detect when I create a service with only `instrumentation_key` in the credentials. It only works if I _also_ have `connection_string` as a credentials key.

I think this is because when you are adding multiple credential keys as 'optional' and not 'required', they need to be in a 'group', i.e. [x,y], so

```
@application.services.find_service(FILTER, CONNECTION_STRING, INSTRUMENTATION_KEY)['credentials']
```
should become
```
@application.services.find_service?(FILTER, [CONNECTION_STRING, INSTRUMENTATION_KEY])
```
to make this an either/or detection. The same for the supports method call to `@application.services.one_service`.

Could you test this and let me know if I'm correct here?

mayrstefan:
Hi,

you are absolutely right. I found the space where I tested my changes and it had a user provided service with both keys so that I could switch between versions 2.+ and 3.+.

Thanks for pointing this out

Issue #986, opened 2022-12-09
Offlinizing java buildpack does not support cflinuxfs4


AnnaAAL:
We are offlinizing the java_buildpack.
We've noticed that when the packaging command for the offline buildpack is executed (as it says in the [README](https://github.com/cloudfoundry/java-buildpack#offline-package)):

**_bundle exec rake clean package OFFLINE=true PINNED=true_** 

In the output it's visible that it's pointing to bionic instead of jammy.

However, on cf push we get the following error: 
_ERROR Finalize failed with exception #<RuntimeError: Jvmkill Agent error: Unable to find cached file for [https://java-buildpack.cloudfoundry.org/jvmkill/jammy/x86_64/index.yml>](https://java-buildpack.cloudfoundry.org/jvmkill/jammy/x86_64/index.yml%3E)
   Jvmkill Agent error: Unable to find cached file for https://java-buildpack.cloudfoundry.org/jvmkill/jammy/x86_64/index.yml
   Failed to compile droplet: Failed to run finalize script: exit status 1_

We've tried locally to edit [package.rb line76](https://github.com/cloudfoundry/java-buildpack/blob/main/rakelib/package.rb#L76) with jammy, i.e. _PLATFORMS = %w[bionic **jammy**].freeze_ and it works. 

Could you, please review our proposal and adjust it/add it to the buildpack, so we can use it offlinized?

dmikusa:
Are you able to access https://java-buildpack.cloudfoundry.org/jvmkill/jammy/x86_64/index.yml? That looks like the correct URL and that loads for me.

AnnaAAL:
Hi @dmikusa , thank you for the prompt feedback. 
We do have access to the given index file, however when we use the given command from the [README](https://github.com/cloudfoundry/java-buildpack#offline-package) :
**bundle exec rake clean package OFFLINE=true PINNED=true**
it does not use jammy, rather than bionic. Here's a screenshot of what's being cached using rake: 
![image](https://user-images.githubusercontent.com/97691430/206988986-9ac491a7-c639-45de-8cd9-6d12df724277.png)

Currently, just for testing, we edited the mentioned line above manually. Could you please recheck and confirm how the buildpack can be offlinized for jammy(cflinuxfs4) using the rake command?

dmikusa:
[...]

PR #991, opened 2023-01-18
Add PLATFORMS var for offline bundling

AnnaAAL:
As discussed in https://github.com/cloudfoundry/java-buildpack/issues/986 there's a need to have PLATFORMS var to allow selecting between bionic and jammy for cflinuxfs3 and cflinuxfs4 stacks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,6 +214,17 @@ Caching https://github.com/SAP/SapMachine/releases/download/sapmachine-11.0.10/s
 Creating build/java-buildpack-offline-cfd6b17.zip
 ```
 
+Different platforms can be selected using the `PLATFORMS` arguments. Supported values of `PLATFORMS` are **bionic** for cflinuxfs3 and **jammy** for cflinuxfs4. The default value is **bionic**.
+
+```bash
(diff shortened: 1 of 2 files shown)


AnnaAAL:
Hi @dmikusa @pivotal-david-osullivan, would you consider this suggestion for fix to #986 ? With this proposal there's no need to compile both bionic and jammy, as the buildpack size is getting larger.

dmikusa:
Thanks for putting this together! I think it's a great solution, and after talking with @pivotal-david-osullivan we can definitely include it.

The feedback we have is that we'd like to see a different default value applied. We'd prefer, for the time being, to have the default be both bionic and jammy.

David also did some testing and found that if the`PLATFORM` env is just empty then the default value is not applied. Checking against `platform_var.empty?` would be a good idea. If it's empty, assign the default.

You might also want to trim whitespace after you split on the `,`. I didn't try it, but if someone enters `bionic, jammy` the extra space might cause issues.

Thanks again for the PR!

pivotal-david-osullivan:
Hi @AnnaAAL, thanks again for the contribution! Are you able to make the adjustments suggested above to handle empty values/whitespace so we can get this merged?

ramonskie:
we refactored the java buildpack to go libbuildpack.
and therefore this PR is invalid.

if this is still necessary please open a new PR

PR #995, opened 2023-02-15
Bring your own Ruby

ryanmoran:
Before the `cflinuxfs4` stack becomes the default in CF, we will be removing the `ruby` interpreter from the set of installed packages. This means that the Java buildpack will need to bring its own interpreter in order to keep running on Ruby.

This PR contains a small `bash` shim that sits in front of each of the buildpack `bin/*` executables. The shim makes sure that there is an installed Ruby interpreter in a predefined location and then `exec`s the equivalent Ruby entrypoint for the buildpack.

The buildpack packaging and configuration have been slightly modified to allow for this to work both online and offline.

The buildpack will run with the latest Ruby 3.1.x version available in the Ruby Buildpack repository.


ryanmoran:
As I was merging a number of other PRs, the tab I had open with this PR got mixed in the shuffle and I accidentally pushed the big green button without thinking. I have already reverted the commit on `main` and I've opened the PR again over here: https://github.com/cloudfoundry/java-buildpack/pull/1000. 😬

PR #1004, opened 2023-03-22
Don't use lsb_release to find platform

ryanmoran:


anthonydahanne:
yeah, fair enough, That makes sense, thanks.

PR #1017, opened 2023-06-02
Install ruby when custom stacks are used

dsabeti:
If a user specifies a custom stack -- for example, a Paketo stack based on Ubuntu Jammy -- this check will fail and ruby will not be installed, causing the app to fail to stage.

This change skips installing ruby _only_ when the specified stack is cflinuxfs3, which is known to include its own ruby.

I considered a few alternatives before landing on this one:
- Removing the check entirely. It would result in an unnecessary installation of ruby for cflinuxfs3, but I suspect it would work all the same.
- Checking for ruby using `which ruby`, as the code does in line 97. In this case, we may want to move the call to `util::environment::setup` into the `if` block:
  ```
  if ! which ruby > /dev/null; then
      mkdir -p "${RUBY_DIR}"

      util::install
      util::environment::setup
    fi
  ```

I'm open to these other solutions, but unsure of their repercussions. I went with the least drastic change.

### DIFF
--- a/bin/run
+++ b/bin/run
@@ -93,7 +93,7 @@ function main() {
   local phase
   phase="$(basename "${0}")"
 
-  if [[ "${CF_STACK:-}" == "cflinuxfs4" ]]; then
+  if [[ "${CF_STACK:-}" != "cflinuxfs3" ]]; then
     if ! which ruby > /dev/null; then
       mkdir -p "${RUBY_DIR}"
 
--- a/bin/run
+++ b/bin/run
@@ -93,15 +93,13 @@ function main() {
   local phase
   phase="$(basename "${0}")"
 
-  if [[ "${CF_STACK:-}" != "cflinuxfs3" ]]; then
-    if ! which ruby > /dev/null; then
-      mkdir -p "${RUBY_DIR}"
+  if ! which ruby > /dev/null; then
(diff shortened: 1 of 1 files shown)


robdimsdale (approved):
This makes sense to me. I think it's better to assume that ruby isn't present on arbitrary stacks so switching the logic around is preferable.

anthonydahanne:
hello @dsabeti !
Actually @pivotal-david-osullivan and I discussed that, and we'd prefer option 2: no check on the stack, but only on the presence of Ruby - so in case cflinuxfs5 brings Ruby back (who knows?!) it will be working fine.

robdimsdale:
As a Paketo/CF stacks maintainer I can state with a high degree of certainty that we won't be bringing ruby back into the stack :)

Checking for the presence of `ruby` is even better than relying on the implicit knowledge of cflinuxfs3 having ruby, but is a bit more work that the existing solution in this PR.

dsabeti:
Cool. I went ahead with option 2. I decided to move the call to `util::environment::setup` inside the same if-block -- the thinking being that, if ruby exists already, those environment variables are probably already set up.

robdimsdale (approved):

anthonydahanne (approved):

pivotal-david-osullivan:
Looks good, thanks all!

PR #1056, opened 2024-02-29
Improvement to condition to trigger java-cfenv framework

pivotal-david-osullivan:
This PR changes the condition that checks when to contribute the Java CFEnv framework for Boot 3 apps. Previously, the condition checks for Spring Boot 3 using either the Manifest entry `Spring-Boot-Version` or presence of `spring-boot-<version>.jar` files.

This change ensures only the presence of the Manifest entry with a version > 3 is used to determine whether the app is a Spring Boot 3 app.


anthonydahanne:
LGTM

PR #1063, opened 2024-03-12
Regex Pattern updated to support both 10.1 and 10.0 Gemfire Tomcat jar file

ainani-vmw:


lib/java_buildpack/container/tomcat/tomcat_geode_store.rb
```diff
@@ -136,8 +136,7 @@ def create_cache_client_xml
 
       def detect_geode_tomcat_version
         geode_tomcat_version = nil
-
-        geode_modules_tomcat_pattern = /ge.*-modules-tomcat(?<version>[0-9]*).*.jar/.freeze
+        geode_modules_tomcat_pattern = /ge.*-tomcat(?<version>[0-9]*).*.jar/.freeze
```
pnikonowicz, line 139: will this break older versions of gemfire / geode ?

ainani-vmw: Yes, it will support both 10.0 and 10.1
Regex pattern verified for both releases.
But will need to run CI using these chnages for 10.0 and then I can confirm if their any issues or not.

pnikonowicz: that makes sense now, thanks. I can't resolve this conversation for some reason. if you're able to, feel free to resolve

onichols-pivotal (approved):

PR #1076, opened 2024-05-16
Adds support for Tomcat 10.1 alongside 9

pivotal-david-osullivan:
Adds support for multiple version lines of Tomcat (similar to JRE lines), as 9 & 10.1 are current stable lines.

### DIFF
--- a/config/packaging.yml
+++ b/config/packaging.yml
@@ -194,9 +194,14 @@ takipi_agent:
   release_notes: '[Release Notes](https://doc.overops.com/docs/whats-new)'
 
 tomcat:
-  name: Tomcat
+  name: Tomcat 9
   cve_notes: '[Security](https://tomcat.apache.org/security-9.html)'
   release_notes: '[ChangeLog](https://tomcat.apache.org/tomcat-9.0-doc/changelog.html)'
+
+tomcat-10:
+  name: Tomcat 10.1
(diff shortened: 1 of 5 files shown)


dmikusa (approved):

anthonydahanne:
Tested this change with this simple webapp: https://github.com/Azure-Samples/tomcat10-jakartaee9

When running it with no specific configuration, it deployed it on `Tomcat Instance 9.0.87` and while no errors were displayed, it could not display (404 error) the `HelloServlet` which relies on `jakarta` while Tomcat 9 does not support `jakarta`

Then, I repushed with:

```yaml
---
applications:
- name: tomcat10
  memory: 2G
  random-route: true
  path: target/helloworld.war
  buildpacks:
    - https://github.com/cloudfoundry/java-buildpack.git#tomcat_version_lines

  env:
    JBP_LOG_LEVEL: info
    JBP_CONFIG_SPRING_AUTO_RECONFIGURATION: '{enabled: false}'
    JBP_CONFIG_OPEN_JDK_JRE: '{ jre: { version: 17.+ } }'
    JBP_CONFIG_TOMCAT: '{ tomcat: { version: 10.1.+ }}'
```

and the servlet could be displayed properly, as I could see `Tomcat Instance 10.1.24 ` was being downloaded and used.

anthonydahanne:
LGTM

PR #1094, opened 2024-11-08
Add multi-tech support to Dynatrace OneAgent integration

arthfl:
This adds handling for a new, optional, property named `addtechnologies` a user can configure in a CF service which enables them to add additional OneAgent code-modules to a given java-buildpack deployment.
This is required in multi-buildpack and sidecar scenarious to also monitor other technologies besides java.

Also had to refactor the URL sanitization a bit, because it was eating up multiple `include` parameters which resulted in displaying a faulty log output.


TimGerlach:
Hi @anthonydahanne,
Can you or someone from the buildpack team please have a look at this PR. We at SAP urgently require this feature.

pivotal-david-osullivan:
Do you have a real SaaS env to test this with? I did not get that far but testing with the following (albeit invalid) values in `addtechnologies` results in the below URL, which is missing the previous tech:

```
      "credentials": {
        "addtechnologies": "node,ruby",
```

```
live.dynatrace.com/api/v1/deployment/installer/agent/unix/paas/latest?include=ruby&bitness=64&Api-Token
```

arthfl:
@pivotal-david-osullivan done :-)

PR #1104, opened 2025-02-27
Update link to SapMachine repository

RealCLanger:
We have moved our web presence to https://sapmachine.io/ so the link needs to be updated. The old URL will still be supported for a while, though.


RealCLanger:
@pivotal-david-osullivan, can you please take care of this?

RealCLanger:
Thanks @pivotal-david-osullivan for pulling this in.

PR #1126, opened 2025-11-03
Update SapMachine JRE version from 11+ to 25+

skateball:
SapMachine 11 reached End of Support in Dec2024. Therefore the default should be changed


skateball:
@RealCLanger my part of CLA should be done since 3Nov, but a company named 'SAP SE' still needs to confirm it. Requested their confirmation again

PR #1220, opened 2026-03-24
fix: normalise tomcat version wildcard + to * (fixes #1219)

ramonskie:
## Summary

Fixes #1219 — staging failed with `improper constraint: 10.1.+` when specifying a two-segment minor version wildcard in `JBP_CONFIG_TOMCAT`.

## Root Cause

`determineTomcatVersion` converted single-segment patterns (`9.+` → `9.x`, `10.+` → `10.x`) but **passed multi-segment patterns through unchanged** (`10.1.+` → `10.1.+`). Neither `blang/semver` nor `Masterminds/semver` accept `+` as a wildcard suffix, so `FindMatchingVersion("10.1.+", ...)` always errored.

The same latent bug affected `10.23.+` — the old test for it asserted the broken pass-through value.

## Fix

Replace `+` with `*` uniformly, consistent with `jre.go`'s `normalizeVersionPattern`:

```go
// before
if !strings.Contains(strings.TrimSuffix(pattern, ".+"), ".") {
    return strings.TrimSuffix(pattern, ".+") + ".x"
}
return pattern  // "10.1.+" passed through unchanged → breaks FindMatchingVersion

// after
return strings.ReplaceAll(match[1], "+", "*")
// "9.+" → "9.*", "10.+" → "10.*", "10.1.+" → "10.1.*", "10.23.+" → "10.23.*"
```

`Masterminds/semver` v1.5.0 treats `x`, `X`, and `*` as equivalent wildcards (`isX()` function), so all patterns resolve correctly.

## Changes

- `src/java/containers/tomcat.go` — simplified `determineTomcatVersion` to a single `strings.ReplaceAll`
- `src/java/containers/tomcat_test.go` — updated expected values (`9.x` → `9.*`, `10.x` → `10.*`, `10.23.+` → `10.23.*`) and added the missing `10.1.+` test case
- `src/integration/tomcat_test.go` — regression integration test using the exact env var combination from the issue report

### DIFF
--- a/src/integration/tomcat_test.go
+++ b/src/integration/tomcat_test.go
@@ -242,6 +242,27 @@ func testTomcat(platform switchblade.Platform, fixtures string) func(*testing.T,
 			})
 		})
 
+		// Regression test for https://github.com/cloudfoundry/java-buildpack/issues/1219
+		// Staging failed with "improper constraint: 10.1.+" when using a two-segment
+		// minor version wildcard (e.g. 10.1.+) in JBP_CONFIG_TOMCAT. The fix normalises
+		// "10.1.+" → "10.1.*" before passing it to libbuildpack's FindMatchingVersion.
+		context("with a two-segment minor version wildcard in JBP_CONFIG_TOMCAT (issue #1219)", func() {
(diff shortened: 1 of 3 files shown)


kiril-keranov (approved):
lgtm

PR #1226, opened 2026-03-27
add support for spring boot 4, with a fallback to 3

zakerf:
related to #1222

### DIFF
--- a/src/java/frameworks/java_cf_env.go
+++ b/src/java/frameworks/java_cf_env.go
@@ -2,10 +2,11 @@ package frameworks
 
 import (
 	"fmt"
-	"github.com/cloudfoundry/java-buildpack/src/java/common"
 	"os"
 	"path/filepath"
 	"strings"
+
+	"github.com/cloudfoundry/java-buildpack/src/java/common"
 
 	"github.com/cloudfoundry/libbuildpack"
 )
@@ -30,14 +31,14 @@ func (j *JavaCfEnvFramework) Detect() (string, error) {
 		return "", nil
 	}
 
-	// Check if Spring Boot 3.x is present
-	if !j.isSpringBoot3() {
+	// Check if Spring Boot 3.x/4.x is present
+	if !j.isSpringBootMajor(4) && !j.isSpringBootMajor(3) {
 		return "", nil
 	}
 
 	// Don't enable if java-cfenv is already in the application
 	if j.hasJavaCfEnv() {
-		j.context.Log.Debug("java-cfenv already present in application")
+		j.context.Log.Info("java-cfenv already present in application")
 		return "", nil
 	}
 
@@ -48,14 +49,27 @@ func (j *JavaCfEnvFramework) Detect() (string, error) {
 func (j *JavaCfEnvFramework) Supply() error {
 	j.context.Log.Debug("Installing Java CF Env")
 
-	// Get java-cfenv dependency from manifest
-	dep, err := j.context.Manifest.DefaultVersion("java-cfenv")
+	dependency := "java-cfenv"
+	defaultVersion := "4.0.0"
+	versionPattern := "4.x.x"
+
+	if j.isSpringBootMajor(3) {
+		defaultVersion = "3.5.1"
+		versionPattern = "3.x.x"
+	}
+
+	allVersions := j.context.Manifest.AllDependencyVersions(dependency)
+	resolvedVersion, err := libbuildpack.FindMatchingVersion(versionPattern, allVersions)
+
+	dep := libbuildpack.Dependency{Name: dependency, Version: resolvedVersion}
 	if err != nil {
-		j.context.Log.Warning("Unable to determine Java CF Env version, using default")
+		j.context.Log.Warning("Unable to determine Java CF Env version for pattern %s, using default", versionPattern)
 		dep = libbuildpack.Dependency{
-			Name:    "java-cfenv",
-			Version: "3.1.0", // Fallback version
+			Name:    dependency,
+			Version: defaultVersion,
 		}
+	} else {
+		j.context.Log.Debug("Resolved Java CF Env version pattern '%s' to %s", versionPattern, resolvedVersion)
 	}
 
 	// Install java-cfenv JAR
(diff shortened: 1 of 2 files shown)


stokpop (changes requested):

zakerf:
@ramonskie the [compatibility matrix](https://github.com/pivotal-cf/java-cfenv?tab=readme-ov-file#compatibility-matrix) also mentions that java-cfenv 4.x is for Spring Boot 4 and java-cfenv 3.x is for Spring Boot 3

manifest.yml
```diff
@@ -283,14 +285,22 @@ dependencies:
   - cflinuxfs4
   source: https://repo1.maven.org/maven2/org/jacoco/jacoco/0.8.14/jacoco-0.8.14.zip
   source_sha256: 0372447f54900b4e77bcb216985b7d31cf5318fc599f8f9346ee35830448c125
-- name: java-cfenv
+- name: java-cfenv-sb3
   version: 3.5.0
   uri: https://java-buildpack.cloudfoundry.org/java-cfenv/java-cfenv-3.5.0.jar
```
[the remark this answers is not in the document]
stokpop: Yep, I saw the size was also an issue in the past when Java 21 was added. There is a lot in the "--cached" buildpack.zip indeed. Did a local check and biggest dependencies are your-kit-profiler (137MB) and jprofiler (119MB). Then we have 5 java versions (8,11,17,21,25) in 3 distributions (OpenJDK, Zulu, SAPMachine) for a total of ~596MB. In comparison: the java-cfenv jar is about 872kb. ;-)

Seems off topic here: maybe for another issue/PR: is it possible to create slimmed down versions? E.g. with only one JDK distribution and with/without profilers?

We create a stripped down offline buildpack ourselves with only what we need by removing entries from the manifest.yml file. (One downside: if you then request something not present, e.g. a profiler that is not included, the cf push fails with an unclear error message).

src/java/frameworks/java_cf_env.go
stokpop, line 42: Might be good to log this at info level to avoid confusion wether cf env is available. Nice to have is to also parse the version in `hasJavaCfEnv` and report the found version.

stokpop (changes requested):

manifest.yml
```diff
@@ -283,14 +285,22 @@ dependencies:
...
   sha256: 6a761fe530783c0ec9e6d1713ef54f6504803bf1ad02856d3ee7b46211f905c5
   cf_stacks:
   - cflinuxfs4
   source: https://repo1.maven.org/maven2/io/pivotal/cfenv/java-cfenv/3.5.0/java-cfenv-3.5.0.jar
   source_sha256: 9cc8d4c368bc90eafb7b6b14bc34c57ef5523f3ec8546e3fbd91326cdfc13500
+- name: java-cfenv
+  version: 4.0.0
+  uri: https://java-buildpack.cloudfoundry.org/java-cfenv/java-cfenv-4.0.0.jar
```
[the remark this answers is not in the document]
zakerf: Should I prep a PR for that or do you have something in the works for it?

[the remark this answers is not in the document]
ramonskie: The manifest will be populated once we add it to the buildpack dependency config in our overall dependency pipeline

ramonskie:
cfenv should be added as a dependency-build and not be added within here in the manifest.
this way we will keep with updates
see https://github.com/cloudfoundry/buildpacks-ci/blob/master/pipelines/dependency-builds

we should add something like there in the config.yml
```
  cfenv:
    buildpacks:
      java:
        lines:
          - line: 3.X.X
          - line: 4.X.X
    source_type: github_releases
    source_params:
      - 'repo: /pivotal-cf/java-cfenv'
      - 'fetch_source: true'
    any_stack: true
    versions_to_keep: 1
 ```
 
 im putting this here in the hope that ppl will understand the structure more clearly and that they could make a pr for this :)

**Comment by zakerf** [comment]
> cfenv should be added as a dependency-build and not be added within here in the manifest. this way we will keep with updates see https://github.com/cloudfoundry/buildpacks-ci/blob/master/pipelines/dependency-builds
> 
> we should add something like there in the config.yml
> 
> ```
>   cfenv:
>     buildpacks:
>       java:
>         lines:
>           - line: 3.X.X
>           - line: 4.X.X
>     source_type: github_releases
>     source_params:
>       - 'repo: /pivotal-cf/java-cfenv'
>       - 'fetch_source: true'
>     any_stack: true
>     versions_to_keep: 1
> ```
> 
> im putting this here in the hope that ppl will understand the structure more clearly and that they could make a pr for this :)

Thanks! I'll prep a PR

**Comment by ramonskie** [comment]
the cfenv has now been added and updated see #1253 and #1251 
so please adjust this pr 
so we can merge it and include it within 5.0.2 release

File: src/java/frameworks/java_cf_env.go
Comment (zakerf) [reply src/java/frameworks/java_cf_env.go]: If an app developer is pulling in the dependency themselves they're probably aware it's present in their app, not sure if it's useful to have an extra log for that?

**Comment by zakerf** [comment]
> the cfenv has now been added and updated see #1253 and #1251 so please adjust this pr so we can merge it and include it within 5.0.2 release

Rebased the branch to latest main to be up to date again

File: src/java/frameworks/java_cf_env.go
Comment (stokpop) [reply src/java/frameworks/java_cf_env.go]: Probably... but there are already teams that have it added as a workaround for until this PR is merged. Might prevent confusion.

File: src/java/frameworks/java_cf_env.go [hunk 3]
```diff
@@ -48,14 +49,27 @@ func (j *JavaCfEnvFramework) Detect() (string, error) {
...
 	j.context.Log.Debug("Installing Java CF Env")
 
-	// Get java-cfenv dependency from manifest
-	dep, err := j.context.Manifest.DefaultVersion("java-cfenv")
+	dependency := "java-cfenv"
+	defaultVersion := "4.0.0"
+	versionPattern := "4.x.x"
+
+	if j.isSpringBootMajor(3) {
+		defaultVersion = "3.1.0"
```
stokpop: should default be java-cfenv 3.5.1?

stokpop (commented):
Check default version for SB3.

src/java/frameworks/java_cf_env.go
zakerf: I think you're right, bumped it to 3.5.1

src/java/frameworks/java_cf_env.go
zakerf: updated it to an info log line

ramonskie:
i do not like the hardcoded defaults. but for now it is okay is i do not expect it to change anytime soon

zakerf:
@ramonskie thanks for the support and merging the pr!

PR #1258, opened 2026-04-29
Issue 1257 fix mem calc vars and update compare doc

### LINKED ISSUE #1257: JBP_CONFIG_*_JRE memory calculator settings ignored
Setting memory calculator tuning parameters via `JBP_CONFIG_OPEN_JDK_JRE` has no effect in the Go buildpack. The configuration appears to be silently ignored.

This is same for all other `JBP_CONFIG_*_JRE` variations.

**Context**

After migrating from the Ruby buildpack to the Go buildpack, apps that ran fine in a 750M container fail at startup with:

```
required memory 1269289K is greater than 750M available for allocation:
  -XX:MaxDirectMemorySize=10M, -Xmx512M, -XX:MaxMetaspaceSize=233001K,
  -XX:ReservedCodeCacheSize=240M, -Xss1M x 250 threads
```

This seems to be caused by the v4 memory calculator now including a user-pinned `-Xmx` in its total memory check (v3 did not) — see also [cloudfoundry/java-buildpack#997](https://github.com/cloudfoundry/java-buildpack/issues/997) for background on the v3→v4 upgrade. A mitigation might be to reduce `stack_threads` via `JBP_CONFIG_OPEN_JDK_JRE` — but this appears to have no effect.

**Expected behaviour**

```yaml
env:
  JBP_CONFIG_OPEN_JDK_JRE: '{ memory_calculator: { stack_threads: 50 } }'
```

Should result in `--thread-count=50` passed to the memory calculator (as documented and as worked in the Ruby buildpack). Reducing from 250 → 50 threads saves 200M of stack, which would make the 750M container viable again.

**Actual behaviour**

The memory calculator always uses the default of 250 threads regardless of the setting.

**Suspected root cause**

Two issues in `src/java/jres/memory_calculator.go`:

1. `LoadConfig()` appears not to be called during `Supply()` or `Finalize()`.
2. Even if called, `LoadConfig()` reads non-standard `MEMORY_CALCULATOR_STACK_THREADS` / `MEMORY_CALCULATOR_HEADROOM` env vars rather than parsing `JBP_CONFIG_OPEN_JDK_JRE`.

**Suggested fix**

- Parse `JBP_CONFIG_OPEN_JDK_JRE` in `LoadConfig()` using the standard CF YAML format.
- Call `LoadConfig()` at the start of **`Supply()`** (before class counting), so both `class_count` and `stack_threads` overrides take effect in time.
- Supported keys: `stack_threads`, `class_count`, `headroom`.

**Related**

- Migration notes should be updated to reflect the memory calculator impact — teams migrating from the Ruby buildpack may be surprised by startup failures.

**Workaround**

None found. `MEMORY_CALCULATOR_STACK_THREADS` might appear to be an alternative env var, but since `LoadConfig()` appears not to be called, it seems equally ignored.

stokpop:
### What changed

**Bug fixes**
- `LoadConfig()` is now called in both `Supply()` and `Finalize()`, so memory calculator settings are applied at staging time (needed for class counting) and at runtime command generation.
- `JBP_CONFIG_OPEN_JDK_JRE` is now parsed for `memory_calculator` settings (`stack_threads`, `headroom`, `class_count`). Previously these were silently ignored.
- `ValidateFields` is scoped to the `memory_calculator` sub-section only, so legitimate top-level keys like `jre:` no longer produce a spurious `WARNING`.

**Improved logging**
- Install log now shows whether each value was user-supplied or auto-detected/default:
  `Memory Calculator installed: Loaded Classes: 23826 (auto-detected), Threads: 250 (default), Headroom: 0% (default)`

**Documentation**
- `RUBY_VS_GO_BUILDPACK_COMPARISON.md`: updated memory calculator row and added migration section explaining the v3 → v4 behaviour change. The difference only affects apps with an explicit `-Xmx` in `JAVA_OPTS` (bad practice in containerised environments). Migration options: lower `stack_threads`, remove `-Xmx`, or increase manifest memory.

**Tests**
- Added `memory_calculator_issues_test.go` covering the above fixes.

### DIFF
--- a/src/java/jres/memory_calculator.go
+++ b/src/java/jres/memory_calculator.go
@@ -19,6 +19,7 @@ type MemoryCalculator struct {
 	jreDir           string
 	jreVersion       string
 	javaMajorVersion int
+	jreName          string
 	calculatorPath   string
 	version          string
 	classCount          int
@@ -31,12 +32,13 @@ type MemoryCalculator struct {
 }
 
 // NewMemoryCalculator creates a new memory calculator
-func NewMemoryCalculator(ctx *common.Context, jreDir, jreVersion string, javaMajorVersion int) *MemoryCalculator {
+func NewMemoryCalculator(ctx *common.Context, jreDir, jreVersion string, javaMajorVersion int, jreName string) *MemoryCalculator {
 	return &MemoryCalculator{
 		ctx:              ctx,
 		jreDir:           jreDir,
 		jreVersion:       jreVersion,
 		javaMajorVersion: javaMajorVersion,
+		jreName:          jreName,
 		stackThreads:     DefaultStackThreads,
 		headroom:         DefaultHeadroom,
 	}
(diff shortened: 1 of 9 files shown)


src/java/jres/memory_calculator.go
```diff
@@ -364,24 +373,73 @@ func (m *MemoryCalculator) convertToRuntimePath(stagingPath string) string {
...
+// (standard CF format) and falls back to MEMORY_CALCULATOR_* env vars.
+// Must be called at the start of Supply(), before countClasses().
 func (m *MemoryCalculator) LoadConfig() {
-	// Check for environment overrides
-	// JBP_CONFIG_OPEN_JDK_JRE='{memory_calculator: {stack_threads: 300}}'
+	if m.configLoaded {
+		return
+	}
+	m.configLoaded = true
+	if config := os.Getenv("JBP_CONFIG_OPEN_JDK_JRE"); config != "" {
```
kiril-keranov: The memory calculator is used with all the other jre types, seems this should be more general to be able to adjust considering the other [jre types](https://github.com/cloudfoundry/java-buildpack/blob/main/src/java/jres/jre.go#L129)

stokpop: Good point, created an update to handle all JBP_CONFIG_*_JRE settings. Also updated the issue: #1257 

I was a bit surprised the difference is there a I expect all JREs to behave similar as they share the open source JRE. There might be differences for Zing and IBM specific J9 implementations though. And maybe per java version. (Also handling the usage of virtual threads might impact the memory calculator settings, something to investigate seperatly.)

Added a follow on issue to look at shared code between JRE's in golang implementation: #1264 

@kiril-keranov not sure if this is correct process: you can resolve this issue if it looks ok now?

kiril-keranov (approved):

src/java/jres/memory_calculator.go
kiril-keranov: looks fine, approved it, but still don't have the rights to merge. @ramonskie can you also check?

PR #1260, opened 2026-04-29
Issue 1259 java opts breaks pre start script

### LINKED ISSUE #1259: pre-start scripts fails when `JAVA_OPTS` contains special characters
When `JAVA_OPTS` is defined in a CF manifest using a YAML block scalar (`>`), CF may deliver the value with literal newlines, or the value may contain pipe characters (`|`) as part of javaagent options:

```yaml
env:
  JAVA_OPTS: >
    -javaagent:$HOME/BOOT-INF/lib/jfr-exporter.jar=enableExecutorMBeans|disableMyFeature
    -XX:+UseZGC
    -XX:+AlwaysPreTouch
    -XX:-ZUncommit
    -XX:MaxDirectMemorySize=256m
```

The `profile.d/00_java_opts.sh` script uses `sed` to substitute `$JAVA_OPTS`, `$HOME` and `$DEPS_DIR` into each `.opts` file. The `sed` delimiter is `|`, and the replacement string is not sanitised, so the command fails when the value contains `|` or newlines:

```
[APP/PROC/WEB/0] OUT Invoking pre-start scripts.
[APP/PROC/WEB/0] ERR sed: -e expression #1, char 257: unterminated `s` command
[APP/PROC/WEB/0] ERR sed: -e expression #1, char 257: unterminated `s` command
[APP/PROC/WEB/0] ERR sed: -e expression #1, char 257: unterminated `s` command
[APP/PROC/WEB/0] ERR sed: -e expression #1, char 257: unterminated `s` command
```

The error repeats once per `.opts` file. Because the script has no `set -e`, execution continues — but all opts content is dropped and the JVM starts with an empty `JAVA_OPTS`: no memory limits, no agents.

Additionally, `&` and `\` in `JAVA_OPTS` are interpreted as sed replacement metacharacters, causing silent data corruption instead of a visible error.

The `ERR` lines are only visible in CF application logs; there is no staging failure or app crash that directly points to this cause.

## Impact

All JVM flags are silently dropped at startup:
- Buildpack-calculated memory settings (`-Xmx`, `-Xss`, `-XX:MaxMetaspaceSize`) absent → OOM / GC thrashing
- Java agents not attached → no APM/tracing instrumentation
- **`-Djava.security.properties` missing** → `CloudFoundryContainerProvider` not registered → CF system certificates not loaded into JVM truststore → TLS handshake failures: `(certificate_unknown) No trusted certificate found`

The last point is tricky to diagnose: the connection between the `sed` error lines at startup and TLS failures later at runtime is not obvious.

## Suggestions

- Avoid `sed` for variable substitution in the assembly script; use bash parameter expansion instead, which has no special-character restrictions
- Normalize `JAVA_OPTS` to a single line before substitution to handle YAML block scalar newlines
- Consider adding `set -e` to the script so that any failure causes a hard stop rather than continuing with an empty or corrupt `JAVA_OPTS`

stokpop:
Fixes #1259

The `profile.d/00_java_opts.sh` script used `sed` to substitute `$JAVA_OPTS`, `$HOME` and `$DEPS_DIR` into `.opts` file content. This breaks at runtime when `JAVA_OPTS` contains:
- **newlines** — from YAML block scalar (`>`) in CF manifest
- **pipe characters** (`|`) — e.g. javaagent options like `agent.jar=featureA|featureB`
- **`&` or `\`** — interpreted as sed replacement metacharacters, causing silent corruption

When `sed` fails, the script continues without `set -e`, silently dropping all opts file content. The JVM starts with an empty `JAVA_OPTS`: no memory limits, no agents.

## Changes

- **Normalize `JAVA_OPTS` to single line** at capture time (`tr` newlines to spaces) before any substitution
- **Replace `sed` with bash parameter expansion** (`${var//find/replace}`) for `$JAVA_OPTS`, `$HOME` and `$DEPS_DIR` substitutions — no special-character restrictions
- **Add regression tests** covering multiline `JAVA_OPTS`, pipe characters, and `$HOME`/`$DEPS_DIR` expansion

## Why `set -e` was not added

`profile.d` scripts are **sourced** by the CF launcher, not run in a subshell — `set -e` would leak to the parent shell. A `kill -TERM $$` guard was also considered, but after analysis it provides little value: with `sed` replaced by bash parameter expansion, there are **no remaining commands in the assembly loop that can fail** (the substitutions are pure bash builtins). A guard checking for empty `JAVA_OPTS` would also be imprecise, since the memory calculator appends to `JAVA_OPTS` via its own separate `java.sh` script *after* `00_java_opts.sh` runs.

### DIFF
--- a/src/java/frameworks/java_opts_writer.go
+++ b/src/java/frameworks/java_opts_writer.go
@@ -93,9 +93,10 @@ if [ -d "$DEPS_DIR/%s/java_opts" ]; then
             opts_content="${opts_content//\$HOME/$HOME}"
             opts_content="${opts_content//\$JAVA_OPTS/$USER_JAVA_OPTS}"
             
-            # Now expand all remaining environment variables using eval with proper escaping
-            # This mimics Ruby buildpack behavior where shell naturally expands variables
-            # Use eval in a subshell to safely expand variables without executing commands
+            # Expand any remaining environment variables in opts content via eval.
+            # Note: eval executes commands, but .opts files are written by the buildpack
+            # at staging time and run within the container context.
+            # This matches how the Ruby buildpack naturally expanded variables via shell.
             opts_content=$(eval "echo \"$opts_content\"")
             
             if [ -n "$opts_content" ]; then
--- a/src/java/frameworks/java_opts_writer.go
+++ b/src/java/frameworks/java_opts_writer.go
@@ -74,7 +74,8 @@ func CreateJavaOptsAssemblyScript(ctx *common.Context) error {
 
 # Save original JAVA_OPTS from environment (user-provided)
 # Normalize to single line: YAML block scalars (>) may introduce newlines
-USER_JAVA_OPTS=$(echo "$JAVA_OPTS" | tr '\n' ' ' | tr -s ' ')
+# xargs trims leading/trailing whitespace and collapses internal spaces
+USER_JAVA_OPTS=$(echo "$JAVA_OPTS" | tr '\n' ' ' | tr -s ' ' | xargs)
 
 # Start building new JAVA_OPTS
 JAVA_OPTS=""
--- a/src/java/frameworks/java_opts_writer.go
+++ b/src/java/frameworks/java_opts_writer.go
@@ -85,15 +85,12 @@ if [ -d "$DEPS_DIR/%s/java_opts" ]; then
             # Read content and expand runtime variables
             opts_content=$(cat "$opts_file")
             
-            # First, expand special variables that need specific handling
-            # Expand $DEPS_DIR variable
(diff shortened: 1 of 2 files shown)


kiril-keranov (approved):

ramonskie:
-  USER_JAVA_OPTS gets a trailing space from the tr '\n' ' ' pipeline. Harmless in most cases, but can cause double-spaces mid-string in assembled opts. Fix: add a trailing whitespace strip
 
-  Pre-existing eval "echo \"$opts_content\"" has a misleading comment claiming it's safe. The comment is factually wrong — eval does execute commands. Not a regression, but the comment should be corrected to accurately describe the trust assumption.

stokpop:
@ramonskie thanks for you feedback, pushed two commits: ready for re-review.

PR #1268, opened 2026-05-11
Issue 1267 fix: jbp config java main

### LINKED ISSUE #1267: JBP_CONFIG_JAVA_MAIN java_main_class and arguments not supported
The Go buildpack ignores `java_main_class` and `arguments` from `JBP_CONFIG_JAVA_MAIN`.

## java_main_class

```yaml
JBP_CONFIG_JAVA_MAIN: '{java_main_class: org.springframework.boot.loader.launch.PropertiesLauncher}'
```

Should override the manifest `Main-Class` and switch to classpath mode (`-cp`). Currently has no effect.

## arguments

```yaml
JBP_CONFIG_JAVA_MAIN: '{arguments: "--custom.arg=value"}'
```

Should append arguments after the main class. Currently has no effect.

Both are supported in the Ruby buildpack via `config/java_main.yml`.

stokpop:
Also added some notes on why the integration test did not find the missing arguments and command line changes.

Fixes #1267


src/java/containers/container.go
```diff
@@ -39,8 +39,26 @@ func (r *Registry) Register(c Container) {
...
 
-// Detect finds the first container that can handle the application
+// Detect finds the first container that can handle the application.
+// If JBP_CONFIG_JAVA_MAIN specifies an explicit java_main_class, the Java Main
+// container is selected unconditionally — before the normal priority order —
+// so it can override higher-priority containers such as Spring Boot.
 func (r *Registry) Detect() (Container, string, error) {
+	cfg := loadJavaMainConfig(r.context.Log)
+	if cfg.JavaMainClass != "" {
+		for _, container := range r.containers {
```
kiril-keranov: seems this loop might be omit if we rely on the current order that the `java main` container is with least priority like currently [here](https://github.com/cloudfoundry/java-buildpack/blob/main/src/java/containers/container.go#L108). If for some reason these container priorities change or a new container is introduced after the `java main` there will be failing tests like `selects Java Main container instead of Spring Boot` signaling this to change, also seems no major container priorities reordering or addition of new containers are expected I guess

kiril-keranov (approved):
looks good, left minor suggestion for possible optimization

src/java/containers/container.go
stokpop: @kiril-keranov Thanks for the review. Assuming the java main is last if available now... if java main not found but there is a configuration line, a warning is given.

PR #1286, opened 2026-05-19
Adjust framework functionalities for the new dependencies paths

kiril-keranov:
After recent dependency updates with regard to automatic PR creation from the dependency pipeline and `cflinuxfs5` adjustments some of the jar dependencies paths and names were changed. Adapting the corresponding frameworks and tests with regard to these changes.
Linked to https://github.com/cloudfoundry/java-buildpack/pull/1281, https://github.com/cloudfoundry/java-buildpack/pull/1278, https://github.com/cloudfoundry/java-buildpack/pull/1273, https://github.com/cloudfoundry/java-buildpack/pull/1271, https://github.com/cloudfoundry/java-buildpack/pull/1272

### DIFF
--- a/src/java/frameworks/azure_application_insights_agent.go
+++ b/src/java/frameworks/azure_application_insights_agent.go
@@ -278,7 +278,7 @@ func (a *AzureApplicationInsightsAgentFramework) getApplicationName() string {
 
 func (a *AzureApplicationInsightsAgentFramework) constructJarPath(agentDir string) error {
 	// Find the installed JAR
-	jarPattern := filepath.Join(agentDir, "applicationinsights-agent-*.jar")
+	jarPattern := filepath.Join(agentDir, "azure-application-insights*.jar")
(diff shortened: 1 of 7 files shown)


ramonskie:
f.jarPath is always an empty string "" at the start of Supply(). os.Stat("") always returns "not found", so the JAR is re-downloaded on every staging — the caching optimization is bypassed entirely.

Fix: Construct the expected path before the existence check (as the original code did), or call constructJarPath first and only download if it fails.

kiril-keranov:
> f.jarPath is always an empty string "" at the start of Supply(). os.Stat("") always returns "not found", so the JAR is re-downloaded on every staging — the caching optimization is bypassed entirely.
> 
> Fix: Construct the expected path before the existence check (as the original code did), or call constructJarPath first and only download if it fails.

This was for the cf-metrics-exporter framework modifications right, I changed it

Issue #1155, opened 2026-01-23
[go-migration] Tomcat version enforced based on used java version.


mshlyukarski:
Based on [Tomcat Compatibility Matrix](https://tomcat.apache.org/whichversion.html), Tomcat version 9x should be able to be used with java versions => 8. This works as expected in the current Ruby based java community buildpack. 

Staging logs example:
```
   -----> Java Buildpack v4.77.0 (offline) | https://github.com/cloudfoundry/java-buildpack.git#1194e2c3
   -----> Downloading Jvmkill Agent 1.17.0_RELEASE from https://java-buildpack.cloudfoundry.org/jvmkill/jammy/x86_64/jvmkill-1.17.0-RELEASE.so (found in cache)
   -----> Downloading Sap Machine JRE 11.0.27_0.0.b0 from https://github.com/SAP/SapMachine/releases/download/sapmachine-11.0.27/sapmachine-jre-11.0.27_linux-x64_bin.tar.gz (found in cache)
   Expanding Sap Machine JRE to .java-buildpack/open_jdk_jre (0.7s)
   JVM DNS caching disabled in lieu of BOSH DNS caching
   -----> Downloading Open JDK Like Memory Calculator 3.13.0_RELEASE from https://java-buildpack.cloudfoundry.org/memory-calculator/jammy/x86_64/memory-calculator-3.13.0-RELEASE.tar.gz (found in cache)
   Loaded Classes: 14797, Threads: 250
   -----> Downloading Client Certificate Mapper 2.0.1 from https://java-buildpack.cloudfoundry.org/client-certificate-mapper/client-certificate-mapper-2.0.1.jar (found in cache)
   -----> Downloading Container Security Provider 1.20.0_RELEASE from https://java-buildpack.cloudfoundry.org/container-security-provider/container-security-provider-1.20.0-RELEASE.jar (found in cache)
   -----> Downloading Tomcat Instance 9.0.104 from https://java-buildpack.cloudfoundry.org/tomcat/tomcat-9.0.104.tar.gz (found in cache)

```

This does not work in such a way in the go based Java buildpack due to the following [check](https://github.com/cloudfoundry/java-buildpack/blob/feature/go-migration/src/java/containers/tomcat.go#L67). Consequently, even though the specific version of java is provided in the manifest
For example:

```
    env:
      JBP_CONFIG_OPEN_JDK_JRE: '{ jre: { version: 11.+ } }'
[...]

Issue #1151, opened 2026-01-21
[go-migration] Compatibility issue with existing apps due to JRE relocation in the Go-based Java Buildpack


ivzhelev:
**Describe the bug**
With the new Go-based Java Buildpack, the location of the JRE has changed: 
for example, from `/home/vcap/app/.java-buildpack/open_jdk_jre` to the new location `/home/vcap/deps/0/jre/jre-17.0.15/`. 
There are applications that invoke Java via a hardcoded relative path inside the droplet: `.java-buildpack/open_jdk_jre/bin/java` so those hardcoded paths no longer exist, which breaks apps and CF tasks that rely on that path. For example, an app used to execute CF Task to execute e.g. the DB schema (so the java is currently being invoked via the full path `.java-buildpack/open_jdk_jre/bin/java`). Or usages like `JAVA_HOME=$PWD/.java-buildpack/open_jdk_jre JAVA_OPTS="-Djava.io.tmpdir=$TMPDIR`

The new Go-based buildpack has a way to introduce `$JAVA_HOME` which will do the job but only for new apps (the existing apps pointing to `/home/vcap/app/.java-buildpack/open_jdk_jre` will fail).

**Reproduction steps**
- Use an app/script that calls Java using the relative path inside the droplet, e.g.:
`.java-buildpack/open_jdk_jre/bin/java -Xms123M org.springframework.boot.loader.launch.JarLauncher`
- Stage the app using the current or the Go-migration branch of the buildpack.
- Run the app or run a CF task that invokes the hardcoded path.
- Observe that the app fails with `No such file or directory`

**Expected behavior**
Existing apps/CF Tasks could continue to run without modification, even if the buildpack internal JRE layout changes.

**Suggested solution**
Introduce **compatibility shims** during staging:
```
export JAVA_HOME=<the-new-java-location> (available in /profile.d/0_java.sh)
mkdir -p /home/vcap/app/.java-buildpack
ln -s $JAVA_HOME /home/vcap/app/.java-buildpack/open_jdk_jre
ln -s $JAVA_HOME /home/vcap/app/.java-buildpack/sap_machine_jre
```
- This preserves `.java-buildpack/open_jdk_jre` and `.java-buildpack/sap_machine_jre` as a symlink to the actual JRE location.
- `$JAVA_HOME` points to the canonical new location.
- Apps with hard-coded paths will continue working.

PR #1288, opened 2026-05-20
refactor: extract BaseJRE to eliminate duplication across standard JRE providers

### LINKED ISSUE #1264: Refactor duplicate JRE implementations into shared BaseJRE struct
Currently the Go buildpack has 6 nearly identical JRE implementations (~1300 lines total): OpenJDK, SapMachine, Zulu, IBM, Oracle, GraalVM. Each has its own copy of `Supply()`, `Finalize()`, `MemoryCalculatorCommand()` etc. with only names/strings differing.

In the Ruby buildpack, this was solved cleanly via inheritance — e.g. `ZuluJRE < OpenJDKLike` was 3 lines.

## Proposed fix

Introduce a shared `BaseJRE` struct with Go embedding:

```go
type BaseJRE struct {
    ctx        *common.Context
    jreName    string
    envVarName string
    memoryCalc *MemoryCalculator
    ...
}

// Supply(), Finalize(), MemoryCalculatorCommand() implemented once on BaseJRE

type ZuluJRE struct { BaseJRE }
type SapMachineJRE struct { BaseJRE }
// etc.
```

This would reduce ~1300 lines to ~200 lines of shared logic + ~5 lines per vendor.

## Context

Discovered while fixing #1257 — the `LoadConfig()` bug (reading only `JBP_CONFIG_OPEN_JDK_JRE` for all JREs) was a direct consequence of the duplicated structure. A minimal fix was applied to pass `jreName` to `NewMemoryCalculator`. The base struct refactor is the structural follow-up.

stokpop:
Closes #1264

## Summary

Reduces ~1300 lines of near-identical code in 6 JRE files to a single `BaseJRE` struct (243 lines) plus thin per-JRE wrappers (~11 lines each).

## Design

`BaseJRE` uses a **template method pattern with injected config/function fields** so implementers never need to override `Supply`/`Finalize`, eliminating the 'forgot to call base' risk from method-override approaches.

**Variation points:**
- `dirPrefixes`/`dirExacts`: drive `findJavaHome` directory matching per JRE
- `extraFinalizeOpts`: hook for JRE-specific JAVA_OPTS (used by IBM JRE: `-Xtune:virtualized -Xshareclasses:none`)
- `installErrNote`: extra context in install error messages (GraalVM: `ensure repository_root is configured`)

`ZingJRE` is intentionally excluded — it has no memory calculator or jvmkill and has genuinely different `Finalize` behaviour.

## Before / After

| | Before | After |
|---|---|---|
| openjdk.go | 254 lines | 11 lines |
| zulu.go | 219 lines | 11 lines |
| sapmachine.go | 219 lines | 11 lines |
| ibm.go | 223 lines | 13 lines |
| oracle.go | 212 lines | 11 lines |
| graalvm.go | 219 lines | 11 lines |
| base_jre.go | — | 243 lines |
| **Total** | **1346 lines** | **311 lines** |

## Tests

Added `standard_jres_test.go` covering all 5 non-OpenJDK standard JREs (OpenJDK already had `openjdk_test.go`). For each JRE:
- `Name()` returns the correct display string
- `Detect()` triggers on the correct env var and not on others
- `Version()` / `JavaHome()` are empty before installation
- `findJavaHome` (via `Finalize()`) locates the JRE-specific subdirectory prefix correctly

All existing and new tests pass.


stokpop:
Added a follow-up commit that improves observability when the JRE `release` file is missing or unreadable:

- `DetermineJavaVersion` now returns an error instead of silently returning 17 when the release file is absent
- `BaseJRE.Finalize()` and `Tomcat.Supply()` both log a `**WARNING**` when falling back to Java 17

This means a broken or incomplete JRE installation will now produce a visible warning in buildpack output rather than silently proceeding with Java 17 assumptions.

ramonskie:
a quick ai review. (with my own context/skilss)
gave me

🔴 Behavioral Change: tomcat.go — version selection now runs unconditionally when JAVA_HOME is set, even if version detection fails
src/java/containers/tomcat.go
Before: If DetermineJavaVersion returned an error (e.g., missing/unreadable release file), the entire Tomcat version selection block was skipped. The code fell through to the manifest default version.
After: On error, javaMajorVersion is silently defaulted to 17, and Tomcat version selection proceeds. The fallback to manifest default (dep.Version == "") is now only reached when JAVA_HOME is not set at all.
Concrete failure scenario: An app using IBM JRE (typically Java 8) where the release file is absent or unreadable. Previously: manifest default Tomcat version used. Now: Tomcat 10.x selected (Java 17 default → javaMajorVersion >= 11 → versionPattern = "10.x"). Tomcat 10.x is Jakarta EE 9+ and is incompatible with Java EE 8 apps. The app will fail at runtime, not at staging.
The javaMajorVersion < 11 guard on line 89 (Tomcat 10.x requires Java 11+) is also bypassed in this scenario because the defaulted version is 17.

🔴 Bug: BaseJRE.Supply() now calls WriteEnvFile, AddBinDependencyLink, and LinkDirectoryInDepDir for ALL JREs — previously only OpenJDK did this
Files: src/java/jres/base_jre.go (new), vs old ibm.go, zulu.go, sapmachine.go, oracle.go, graalvm.go
The old per-JRE implementations of Supply() for IBM, Zulu, SapMachine, Oracle, and GraalVM did not call:
- ctx.Stager.WriteEnvFile("JAVA_HOME", javaHome) — writes JAVA_HOME to the deps env dir for subsequent buildpacks
- ctx.Stager.AddBinDependencyLink(javaBin, "java") — symlinks java binary into the shared bin/ dir (puts it on PATH for subsequent buildpacks)
- ctx.Stager.LinkDirectoryInDepDir(libDir, "lib") — links the JRE lib/ dir into deps (adds native libs to LD_LIBRARY_PATH)
Only openjdk.go had these three calls. BaseJRE.Supply() now applies them to all six JREs.
Impact: This is a behavioral expansion, not just a refactor. For multi-buildpack staging scenarios, Zulu/IBM/Oracle/SapMachine/GraalVM JREs will now expose JAVA_HOME, java on PATH, and native libs to subsequent buildpacks — they did not before. This is likely the correct behavior (it was probably an oversight in the original per-JRE implementations), but it is an undocumented behavioral change that could affect existing multi-buildpack setups if a subsequent buildpack was relying on JAVA_HOME not being set.



🟡 DetermineJavaVersion no longer handles missing release file silently
Previously returned (17, nil) for missing files; now returns (0, error). All callers in this PR handle it, but the safe default is now scattered — future callers could get 0 instead of a safe default if they forget to handle the error.

🟡 Finalize() may use empty b.version
If Finalize() is called without a prior Supply() (multi-buildpack staging model), b.version will be "". Not a regression from the old code, but the consolidation makes it more visible.

🟡 BaseJRE.Finalize() — WriteJavaOpts with base opts now applies to ALL JREs
src/java/jres/base_jre.go
The old openjdk.go Finalize() wrote base JAVA_OPTS (-Djava.io.tmpdir=$TMPDIR -XX:ActiveProcessorCount=$(nproc)). The old IBM, Zulu, SapMachine, Oracle, and GraalVM Finalize() implementations did not write these base opts.
BaseJRE.Finalize() now writes these opts for all JREs. This is likely correct behavior, but it is an undocumented behavioral expansion. For IBM JRE specifically, -XX:ActiveProcessorCount is a HotSpot flag and may not be recognized by IBM J9 JVM, potentially causing a startup warning or error depending on the IBM JRE version.

i' am okay with the things below.
🟢 IBM JRE default Java version changed
Old IBM JRE defaulted to Java 8 on version detection failure; BaseJRE now defaults to 17 for all JREs. Low severity since IBM JRE is rarely used, but it's a silent behavioral change.

i will still do a full review later on

stokpop:
Thanks for the review! @ramonskie can you have another view? 

A general note on approach: we tried to match the Ruby 4.x buildpack behaviour as the baseline, only deviating where there are concrete issues or PR comments that justify it.

**1. tomcat.go — version detection failure (commit 2c5754bd)**

The Ruby buildpack has no Java-version-aware Tomcat selection — it always defers to the manifest default (`9.+`). I think on detection failure the safe fallback is `DefaultVersion("tomcat")` rather than assuming Java 17 → Tomcat 10.x (see #1155). An explicitly configured version (e.g. `JBP_CONFIG_TOMCAT: '{tomcat: {version: "9.+"}}'`) is still always honoured.

**2. WriteEnvFile / AddBinDependencyLink / LinkDirectoryInDepDir for all JREs**

I assume this is fine as-is. Ruby never set JAVA_HOME at all — setting it for all JREs is intentional new behaviour in the Go migration (see #1151). I don't think a per-JRE flag is needed here, but open to arguments if there's a specific case where it causes problems.

**3. -XX:ActiveProcessorCount for all JREs (commit bf92b625)**

In Ruby `open_jdk_like_jre.rb` adds it for HotSpot only; `ibm_jre_initializer.rb` never does. Moved it out of `baseOpts` and into `extraFinalizeOpts` on each HotSpot JRE (OpenJDK, Oracle, SapMachine, Zulu, GraalVM). IBM keeps its existing opts unchanged.

**4. DetermineJavaVersion returns 0 on error (commit 40e44905)**

Returns `17` on all error paths so callers that don't check the error still get a safe value.

Also added some notes in the migration md so things that do change do not go unnoticed.

ramonskie (approved):
lgtm

stokpop:
@kiril-keranov can you have a look please for a review? thanks

tnikolova82 (approved):
lgtm

stokpop:
@tnikolova82 thanks!

kiril-keranov (commented):
lgtm, was unavailable for a couple of days so pinged Tedi to approve

PR #1316, opened 2026-06-10
fix: replace eval-based start command with shell-free javaexec launcher (#1301)

### LINKED ISSUE #1301: v5.x regression: `JAVA_OPTS` handling broken (pipes/newlines in 5.0.2, quoting in 5.0.3)
## Summary  
There are regressions in `JAVA_OPTS` handling across the v5.x line:  
- **v5.0.2**: breaks when `JAVA_OPTS` contains `|` or newlines (as per #1259)  
- **v5.0.3**: fixes this but introduces a new issue where **quoted JVM args are broken**  
Both behaviours were **not present in v4.76.0**.

---

## Root Cause  
- **5.0.2 (#1259)**: `sed` substitution fails with `|` / newlines → JVM options may be dropped or corrupted  
- **5.0.3** introduces:  
  ```bash
  USER_JAVA_OPTS=$(echo "$JAVA_OPTS" | tr '\n' ' ' | tr -s ' ' | xargs)
  ```  
  → `xargs` removes quoting → argument boundaries lost → downstream shell parsing splits args and may expand `*`

---

## Minimal Reproducer  
```bash
JAVA_OPTS='-Dfoo="bar baz" -DcronSched="0 */7 * * * *"'
USER_JAVA_OPTS=$(echo "$JAVA_OPTS" | tr '\n' ' ' | tr -s ' ' | xargs)
set -f; eval set -- $USER_JAVA_OPTS; printf '<%s>\n' "$@"
# => quoted args split; cron expression becomes multiple tokens
<-Dfoo=bar>
<baz>
<-DcronSched=0>
<*/7>
<*>
<*>
<*>
<*>
```

---

## Real CF Failure  
```text
   2026-05-26T23:15:42.56+0000 [APP/PROC/WEB/0] OUT Invoking pre-start scripts.
   2026-05-26T23:15:42.58+0000 [APP/PROC/WEB/0] OUT Created symlink: /home/vcap/app/BOOT-INF/lib/client-certificate-mapper-2.0.1.jar -> /home/vcap/deps/2/client_certificate_mapper/client-certificate-mapper-2.0.1.jar
   2026-05-26T23:15:42.58+0000 [APP/PROC/WEB/0] OUT Invoking start command.
   2026-05-26T23:15:42.58+0000 [APP/PROC/WEB/0] OUT JVM Memory Configuration: -XX:MaxDirectMemorySize=10M -XX:MaxMetaspaceSize=169241K -XX:ReservedCodeCacheSize=240M -Xss1M -Xmx576230K
   2026-05-26T23:15:42.63+0000 [APP/PROC/WEB/0] ERR Error: Could not find or load main class *.1
   2026-05-26T23:15:42.63+0000 [APP/PROC/WEB/0] ERR Caused by: java.lang.ClassNotFoundException: *.1
   2026-05-26T23:15:42.64+0000 [API/0] OUT Process became ready with guid fe7a5b4a-fba8-48cb-8283-c5a8d1f61132 payload: {"instance"=>"8dcab366-fa64-4dcc-4bad-5f7d", "index"=>0, "cell_id"=>"0b0ac42c-af1a-4963-85a2-dabc9f16253c", "ready"=>true, "version"=>"57f8bb8d-5b81-4b64-8047-c79ab1b62a35"}
   2026-05-26T23:15:42.64+0000 [APP/PROC/WEB/0] OUT Exit status 1
```
`Could not find or load main class *.1` is due to the quotes being stripped from `-DcronSched="0 */1 * * * *"`

---

## Impact  
Breaks valid JVM usage such as:
```text
-Dkey="value with spaces"
-DcronSched="0 */7 * * * *"
```
This is not cron-specific — affects any quoted JVM argument.

---

## Expected  
- `JAVA_OPTS` should preserve argument boundaries  
- special characters (`|`, `*`, whitespace) must not corrupt or expand arguments

stokpop:
## Problem

The start command used `eval "exec java $JAVA_OPTS ..."` to launch the JVM.
`eval` reinterprets its argument as shell code, which made it fundamentally
difficult to preserve JAVA_OPTS values exactly as specified:

- Backslashes were consumed as escape characters (`C:\path` → `C:path`)
- Quotes were stripped, so space-containing values split into multiple
  JVM arguments (`-Dfoo="bar baz"` → `-Dfoo=bar` + `baz`)
- Glob characters in unquoted values expanded (`*/7` → filenames)
- Pipe and semicolon characters were interpreted as shell operators

Each issue required a new escaping workaround; combinations of edge cases
kept producing regressions. The root cause is that `eval` is the wrong tool
for passing an arbitrary string as JVM arguments.

Note: this PR **replaces** #1303 (which can be closed if this is merged, or revived from draft if this PR is not to be merged).

## Solution

**1.** `javaexec` launcher (`src/java/javaexec/`)

A compiled Go binary invoked as `exec $DEPS_DIR/<idx>/bin/javaexec "$JAVA_HOME/bin/java" <args>`.
It tokenizes `$JAVA_OPTS` using POSIX word-splitting and quote-removal rules,
but performs **no further processing**: `$(...)`, `${...}`, backtick
substitutions, and shell metacharacters (`&`, `|`, `;`, `>`) are all passed
literally to the JVM. Nested `$(...)` and `${...}` forms with spaces inside
are kept as one token.

**2.** Pure-bash `_expand_env_vars` in `profile.d/00_java_opts.sh`

Expands `$VAR` / `${VAR}` references using only bash builtins — never
re-interprets quotes, backslashes, or metacharacters. The single known trusted
substitution `$(nproc)` is resolved explicitly. `\$VAR` passes a literal
`$VAR` to the JVM (Ruby buildpack parity). A WARNING is emitted to stderr
(showing only the offending token) when `$(...)` or a backtick appears in
user `JAVA_OPTS`.

**Other fixes included**

- `memory_calculator.go`, `jvmkill.go`: replace hardcoded `/home/vcap/deps/`
  with `$DEPS_DIR` so installs work on non-default CF stacks.

## Test coverage

- Tokenizer unit tests: quoting, metacharacters, `$(...)` grouping,
  `${...}` extended forms, backtick, nested parens, full manifest scenario.
- End-to-end `runStartCommand` tests: exact #1301 reproducer (cron + quoted
  spaces), pipe, metacharacters, command substitution not executed (marker
  file guard), `\$VAR` escape, warning token extraction, POSIX word-split
  documentation.
- Integration: `runStartCommand` smoke test verifying javaexec fires for
  Spring Boot, Java Main, Play, Tomcat containers.

Closes #1301

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ tmp/
 .cache/
 bin/detect
 bin/finalize
+bin/javaexec
 bin/release
 bin/supply
 /*.md
--- a/bin/finalize
+++ b/bin/finalize
@@ -14,6 +14,8 @@ output_dir=$(mktemp -d -t finalizeXXX)
 pushd $BUILDPACK_DIR > /dev/null
 echo "-----> Running go build finalize"
 GOROOT=$GoInstallDir $GoInstallDir/bin/go build -mod=vendor -o $output_dir/finalize ./src/java/finalize/cli
+echo "-----> Running go build javaexec"
+GOROOT=$GoInstallDir $GoInstallDir/bin/go build -mod=vendor -o $output_dir/javaexec ./src/java/javaexec/cli
 popd > /dev/null
 
-$output_dir/finalize "$BUILD_DIR" "$CACHE_DIR" "$DEPS_DIR" "$DEPS_IDX" "$PROFILE_DIR"
+JAVAEXEC_BINARY_PATH=$output_dir/javaexec $output_dir/finalize "$BUILD_DIR" "$CACHE_DIR" "$DEPS_DIR" "$DEPS_IDX" "$PROFILE_DIR"
--- a/docs/DEVELOPING.md
+++ b/docs/DEVELOPING.md
@@ -97,6 +97,7 @@ Build the buildpack binaries:
 This creates executables in the `bin/` directory:
 - `bin/supply` - Staging phase binary (downloads and installs dependencies)
 - `bin/finalize` - Finalization phase binary (configures runtime)
+- `bin/javaexec` - Shell-free JVM launcher (tokenizes `JAVA_OPTS` without `eval`)
 
 ## Project Structure
 
@@ -118,6 +119,7 @@ java-buildpack/
 │   ├── jres/                           # JRE implementations (7 providers)
 │   ├── supply/cli/                     # Supply phase entrypoint
 │   ├── finalize/cli/                   # Finalize phase entrypoint
+│   ├── javaexec/cli/                   # Shell-free JVM launcher entrypoint
 │   ├── resources/                      # Resource configuration files
 │   └── integration/                    # Integration tests
 ├── scripts/                            # Build and test scripts
@@ -156,6 +158,7 @@ Build for the default platform (Linux):
 ```
 -----> Building supply for linux
 -----> Building finalize for linux
+-----> Building javaexec for linux
 -----> Build complete
 ```
 
@@ -193,6 +196,25 @@ go build -mod vendor -o bin/supply src/java/supply/cli/main.go
 
 # Build finalize
 go build -mod vendor -o bin/finalize src/java/finalize/cli/main.go
+
+# Build javaexec (shell-free JVM launcher, required at runtime)
+go build -mod vendor -o bin/javaexec src/java/javaexec/cli/main.go
+```
+
+### Source/Git Buildpack Usage
+
+When deploying with a git URL (`cf push -b https://github.com/.../java-buildpack.git`),
+Cloud Foundry runs `bin/finalize` directly from the cloned source. In that mode
+`bin/javaexec` does not exist (only packaged buildpacks have it). `bin/finalize`
+therefore builds `javaexec` into a temp directory alongside the finalize binary
+and passes the path via the `JAVAEXEC_BINARY_PATH` environment variable.
+`InstallJavaexecLauncher()` prefers this override and falls back to
+`bin/javaexec` for packaged buildpacks.
+
+To verify this path locally (no CF required):
+
+```bash
+bash scripts/test-javaexec-source-path.sh
 ```
 
 ## Running Tests
--- a/docs/design.md
+++ b/docs/design.md
@@ -186,6 +186,12 @@ During the finalize phase (`bin/finalize`), the buildpack:
 1. **Finalizes JRE**: Configures JVM options, memory calculator
 2. **Finalizes Frameworks**: Adds agent paths, system properties
 3. **Finalizes Container**: Generates launch command
+4. **Installs `javaexec`**: Copies the shell-free JVM launcher into
+   `$DEPS_DIR/<idx>/bin/javaexec`. The generated start command invokes it
+   instead of `eval "exec java $JAVA_OPTS ..."` so that `JAVA_OPTS` is
+   tokenized without shell interpretation. For source/git buildpack usage
+   `bin/finalize` builds `javaexec` on the fly and passes its path via
+   `JAVAEXEC_BINARY_PATH` (see [DEVELOPING.md](DEVELOPING.md#sourcegit-buildpack-usage)).
 
 Components can:
 - Read installed dependencies from `$DEPS_DIR/<idx>/`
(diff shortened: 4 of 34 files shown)


ramonskie:
**Medium:** `bin/javaexec` is required but not built for git/source buildpack usage.

`installJavaexecLauncher()` now reads `<buildpackDir>/bin/javaexec`, which works for packaged buildpacks because `scripts/build.sh` creates it. But direct git/source usage only builds the finalize binary via `bin/finalize`; it does not build `src/java/javaexec/cli`, and `bin/javaexec` is not checked in.

This means source buildpack usage such as `WithBuildpacks("https://github.com/cloudfoundry/java-buildpack.git")` can fail during finalize with “javaexec launcher binary not found”.

Suggested fix: update the source/git wrapper to build `javaexec` too, or make `installJavaexecLauncher()` use a generated/temp `javaexec` path when running from source.

ramonskie:
I did a deeper pass on the `javaexec` integration. The launcher approach itself looks like the strongest option for the problem this PR is solving: bash arrays, `JAVA_TOOL_OPTIONS`/`JDK_JAVA_OPTIONS`, generated shell wrappers, or per-container fixes either reintroduce a parser/eval problem, do not preserve current `JAVA_OPTS` semantics/order reliably, or only address part of the eval surface.

The issue I think still needs addressing is lifecycle/packaging integration:

- `installJavaexecLauncher()` runs unconditionally during finalize.
- It requires `<buildpackDir>/bin/javaexec` to exist.
- Packaged buildpacks get that from `scripts/build.sh`, but source/git usage via `bin/finalize` only builds `src/java/finalize/cli` into a temp dir.
- Because `bin/javaexec` is ignored/not checked in, source/git usage can fail staging for any app type, including apps whose final start command does not use `javaexec`, since finalize aborts before release completes.

`WriteProfileD` itself does not look directly broken. `CreateJavaOptsAssemblyScript()` still writes `profile.d/00_java_opts.sh` before the launcher install step. The failure mode is that profile.d may be written successfully, then finalize aborts while copying the missing launcher.

A safer fix would be to make the source/git wrapper build `src/java/javaexec/cli` into the same temp directory as the finalize binary, then pass that explicit binary path into finalize, for example via an env var. `installJavaexecLauncher()` could prefer that override and fall back to `<buildpackDir>/bin/javaexec` for packaged buildpacks. That keeps packaged behavior unchanged, avoids mutating the source checkout during staging, and preserves the good part of this PR: shell-free handling of user-provided `JAVA_OPTS` at the final JVM exec.

stokpop:
Thanks for the thorough review. Addressed in the three commits just pushed:

**`fix(finalize): build javaexec in bin/finalize for source/git buildpack usage`**

- `bin/finalize` now builds `src/java/javaexec/cli` into the same temp dir as the finalize binary and passes the path via `JAVAEXEC_BINARY_PATH`.
- `InstallJavaexecLauncher()` (renamed from private) prefers `JAVAEXEC_BINARY_PATH`, falls back to `<buildpackDir>/bin/javaexec` for packaged buildpacks — no change to packaged behaviour.
- Three unit tests: packaged path, env-var override (source/git path), missing-binary error.

**`docs: document javaexec binary and source/git buildpack path`**

- `DEVELOPING.md`: `javaexec` added to build output, project structure, manual build section, and a new _Source/Git Buildpack Usage_ section explaining the `JAVAEXEC_BINARY_PATH` mechanism.
- `design.md`: finalize phase step for javaexec install + source/git note.
- `src/integration/README.md`: explicit note that source/git path is not covered by integration tests, pointing to unit tests and the smoke script.

**`test(scripts): add JAVA_OPTS tokenization smoke test to source-path script`**

`scripts/test-javaexec-source-path.sh` — verifies the full source/git path locally without CF: builds both binaries, runs the unit tests with the env-var override, then invokes the real `javaexec` binary against a fake JVM to confirm tokenization (quoted spaces, cron `*`, `$(...)` not executed). Useful when `custom_buildpacks` is disabled on the target CF.

stokpop:
@ramonskie thanks for bringing the source/git use case to my attention, I was not aware of that. 
Can you please have another look?

ramonskie:
could you squash the commit or if needed squash it in sensible parts..

stokpop:
@ramonskie cleaned up the commits

ramonskie (approved):
lgtm

PR #1358, opened 2026-07-10
Fix pointing exact version with BP_JAVA_VERSION

### LINKED ISSUE #1338: BP_JAVA_VERSION with exact patch version (e.g. 17.0.13) silently fails
When `BP_JAVA_VERSION` is set to an exact patch version like `17.0.13`, the Go buildpack does not resolve it correctly.

**Root cause:** `normalizeVersionPattern()` in `src/java/jres/jre.go` appends `.*` to any string that contains neither `+` nor `*`, so `"17.0.13"` becomes `"17.0.13.*"`. `libbuildpack.FindMatchingVersion` finds no match for a 4-part wildcard against a 3-part version string, causing an error.

**Ruby buildpack behaviour:** `17.0.13` was treated as an exact pin and resolved directly.

**Affected patterns:**

| Input | Normalized | Result |
|-------|-----------|--------|
| `17` | `17.*` | ✅ works |
| `17.*` | `17.*` | ✅ works |
| `17.0.+` | `17.0.*` | ✅ works |
| `17.0.13` | `17.0.13.*` | ❌ no match → error |

**Fix:** In `normalizeVersionPattern`, detect a fully-qualified version (e.g. 3-part semver with no wildcards) and return it unchanged, or append `.*` only when the version has fewer than 3 parts.

kiril-keranov:
Fixes #1338 
When specifying an exact version for `BP_JAVA_VERSION`, staging fails:
```
2026-07-10T17:10:30.21+0300 [STG/0] OUT **WARNING** Could not find openjdk matching version 17.0.19.*: improper constraint: 17.0.19.*
2026-07-10T17:10:30.21+0300 [STG/0] OUT **ERROR** Failed to install JRE: failed to determine OpenJDK version from manifest: no version of openjdk matching 17.0.19.* found
```

### DIFF
--- a/src/java/jres/jre.go
+++ b/src/java/jres/jre.go
@@ -311,11 +311,17 @@ func GetJREVersion(ctx *common.Context, jreName string) (libbuildpack.Dependency
 }
 
 func normalizeVersionPattern(version string) string {
-	if strings.Contains(version, "+") {
-		return strings.ReplaceAll(version, "+", "*")
-	}
 	if strings.Contains(version, "*") {
 		return version
+	}
+	// Exact version with build metadata (e.g. "17.0.19+11") — pass through as-is.
+	// Must be checked before the general "+" replacement below.
+	exactVersionWithBuildRegex := regexp.MustCompile(`^\d+\.\d+\.\d+\+\d+$`)
+	if exactVersionWithBuildRegex.MatchString(version) {
+		return version
+	}
+	if strings.Contains(version, "+") {
+		return strings.ReplaceAll(version, "+", "*")
 	}
 	// Exact patch version (e.g. "17.0.13") — already fully specified, don't append ".*"
 	// which would produce an unmatchable pattern like "17.0.13.*".
--- a/src/java/jres/jre.go
+++ b/src/java/jres/jre.go
@@ -310,14 +310,16 @@ func GetJREVersion(ctx *common.Context, jreName string) (libbuildpack.Dependency
 	return dep, nil
 }
 
+var exactVersionWithBuildRegex = regexp.MustCompile(`^\d+\.\d+\.\d+\+\d+$`)
+var exactVersionRegex = regexp.MustCompile(`^\d+\.\d+\.\d+$`)
+
 func normalizeVersionPattern(version string) string {
 	if strings.Contains(version, "*") {
 		return version
 	}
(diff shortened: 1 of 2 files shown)


src/java/jres/jre.go
```diff
@@ -317,6 +317,12 @@ func normalizeVersionPattern(version string) string {
 	if strings.Contains(version, "*") {
 		return version
 	}
+	// Exact patch version (e.g. "17.0.13") — already fully specified, don't append ".*"
+	// which would produce an unmatchable pattern like "17.0.13.*".
+	exactVersionRegex := regexp.MustCompile(`^\d+\.\d+\.\d+$`)
```
stokpop: Although this code is probably not on a hot path, you can also hoist regexes so its no compiled for each call, e.g.

```
var threePartVersionPattern = regexp.MustCompile(`^\d+\.\d+\.\d+$`)

func isValidVersion(version string) bool {
	return threePartVersionPattern.MatchString(version)
}
```

src/java/jres/jre_test.go
```diff
@@ -313,6 +313,14 @@ dependencies:
 				Expect(dep.Name).To(Equal("openjdk"))
 				Expect(dep.Version).To(Equal("17.0.13"))
 			})
+
+			It("resolves exact patch version", func() {
+				os.Setenv("BP_JAVA_VERSION", "17.0.13")
```
stokpop: The basic fix probably works as far as I can tell, but the actual versions of JRE are like: X.Y.Z+W format (17.0.19+11). Consider adding a test where BP_JAVA_VERSION resolves against a manifest entry 17.0.19+11. 

I also thought we use "JBP_CONFIG_OPEN_JDK_JRE: '{ jre: { version: "25.0.3" } }'", I assume the fix also works for those cases, but maybe also add test case to be sure.

stokpop (changes requested):
Thanks for the PR! Have made some notes.

src/java/jres/jre.go
kiril-keranov: extracted as proposed

src/java/jres/jre_test.go
kiril-keranov: added also a test for `JBP_CONFIG_OPEN_JDK_JRE: '{ jre: {version: "17.0.19+11"} }'"`

stokpop (approved):
Thanks for the changes

PR #1360, opened 2026-07-15
Add Elastic Otel Java agent

jackshirazi:
We have customers requesting our agent to be added to cloudfoundry. I've used the existing Elastic APM agent as a template

### DIFF
--- a/src/integration/frameworks_test.go
+++ b/src/integration/frameworks_test.go
@@ -400,6 +400,45 @@ func testFrameworks(platform switchblade.Platform, fixtures string) func(*testin
 				})
 			})
 
+			context("with Elastic OTel service binding", func() {
+				it("detects and installs Elastic OTel Java agent", func() {
+					deployment, logs, err := platform.Deploy.
+						WithServices(map[string]switchblade.Service{
+							"elastic-otel": {
+								"otel.exporter.otlp.endpoint": "https://elastic-otel.example.com:443",
+								"api_key":                     "test-api-key",
+							},
+						}).
+						WithEnv(map[string]string{
+							"BP_JAVA_VERSION": "17",
+						}).
+						Execute(name, filepath.Join(fixtures, "apps", "integration_valid"))
+					Expect(err).NotTo(HaveOccurred(), logs.String)
+
+					Expect(logs.String()).To(ContainSubstring("Elastic OTel"))
+					Eventually(deployment).Should(matchers.Serve(ContainSubstring("")))
+				})
+
+				it("configures Elastic OTel with explicit OTLP headers", func() {
+					deployment, logs, err := platform.Deploy.
+						WithServices(map[string]switchblade.Service{
+							"my-elastic-otel": {
+								"otel.exporter.otlp.endpoint": "https://elastic-otel.production.example.com:443",
+								"otel.exporter.otlp.headers":  "Authorization=ApiKey explicit-key",
+								"otel.service.name":           "my-java-app",
+							},
+						}).
+						WithEnv(map[string]string{
+							"BP_JAVA_VERSION": "17",
+						}).
+						Execute(name, filepath.Join(fixtures, "containers", "spring_boot_staged"))
+					Expect(err).NotTo(HaveOccurred(), logs.String)
+
+					Expect(logs.String()).To(ContainSubstring("Elastic OTel"))
+					Eventually(deployment).Should(matchers.Serve(ContainSubstring("")))
+				})
+			})
+
 			context("with OpenTelemetry service binding", func() {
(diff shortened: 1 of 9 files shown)


jackshirazi:
@kiril-keranov @stokpop you seem to be active maintainers, could you please tell me what I need to do to progress this PR

src/java/frameworks/elastic_otel_java_agent.go
```diff
@@ -0,0 +1,263 @@
...
+	return &ElasticOtelJavaAgentFramework{context: ctx}
+}
+
+// Detect checks if the Elastic OTel Java agent should be enabled.
+func (e *ElasticOtelJavaAgentFramework) Detect() (string, error) {
+	if os.Getenv("ELASTIC_OTEL_AGENT") != "" {
+		e.context.Log.Debug("Elastic OTel Java agent framework detected via ELASTIC_OTEL_AGENT")
+		return "elastic-otel-javaagent", nil
+	}
+	if os.Getenv("OTEL_EXPORTER_OTLP_ENDPOINT") != "" && os.Getenv("OTEL_EXPORTER_OTLP_HEADERS") != "" {
```
kiril-keranov: Both `ElasticOtelJavaAgentFramework` and `SplunkOtelJavaAgentFramework` activate when `OTEL_EXPORTER_OTLP_ENDPOINT + OTEL_EXPORTER_OTLP_HEADERS` are set. Splunk fires on `OTEL_EXPORTER_OTLP_ENDPOINT` alone, so setting both env vars installs two -javaagent flags into the same JVM. Can this cause potentially duplicate telemetry and classloader conflicts?

src/java/frameworks/elastic_otel_java_agent.go
```diff
@@ -0,0 +1,263 @@
...
+
+	e.context.Log.BeginStep("Configuring Elastic OTel Java agent")
+
+	relPath, err := filepath.Rel(e.context.Stager.DepDir(), e.jarPath)
+	if err != nil {
+		return fmt.Errorf("failed to determine relative path for Elastic OTel Java agent: %w", err)
+	}
+	runtimeJarPath := filepath.Join(fmt.Sprintf("$DEPS_DIR/%s", e.context.Stager.DepsIdx()), relPath)
+
+	config := e.buildConfiguration()
```
kiril-keranov: When `ELASTIC_OTEL_AGENT=true` triggers detection with no service binding and no `OTEL_EXPORTER_OTLP_HEADERS`, `buildConfiguration()` produces no OTLP auth headers. The agent starts and silently drops all telemetry. The `hasRequiredCredentials` guard only applies on the service-binding path.

src/java/frameworks/elastic_otel_java_agent.go
```diff
@@ -0,0 +1,263 @@
...
+	}
+
+	if endpoint := os.Getenv("OTEL_EXPORTER_OTLP_ENDPOINT"); endpoint != "" {
+		config["otel.exporter.otlp.endpoint"] = endpoint
+	}
+	if headers := os.Getenv("OTEL_EXPORTER_OTLP_HEADERS"); headers != "" {
+		config["otel.exporter.otlp.headers"] = headers
+	}
+	if serviceName := os.Getenv("OTEL_SERVICE_NAME"); serviceName != "" {
+		config["otel.service.name"] = serviceName
```
kiril-keranov: The credential loop on line 153 can write `otel.service.name` from the service binding, but `OTEL_SERVICE_NAME` unconditionally overwrites it with no guard and no log warning. The fallback to `GetApplicationName` does have a !ok guard, but `OTEL_SERVICE_NAME` does not — so env silently beats the service credential.

src/integration/frameworks_test.go
```diff
@@ -400,6 +400,45 @@ func testFrameworks(platform switchblade.Platform, fixtures string) func(*testin
...
+							},
+						}).
+						WithEnv(map[string]string{
+							"BP_JAVA_VERSION": "17",
+						}).
+						Execute(name, filepath.Join(fixtures, "apps", "integration_valid"))
+					Expect(err).NotTo(HaveOccurred(), logs.String)
+
+					Expect(logs.String()).To(ContainSubstring("Elastic OTel"))
+					Eventually(deployment).Should(matchers.Serve(ContainSubstring("")))
```
kiril-keranov: Both Elastic OTel integration tests use `matchers.Serve(ContainSubstring(""))`, which matches any response including error pages or empty bodies. A broken agent that crashes the app seems to still pass these tests.

kiril-keranov (commented):
Thanks for the contribution. Left some comments on findings

PR #1397, opened 2026-08-20
Refactor supply and fix possible issues in frameworks

kiril-keranov:
- Refactor `supply.go` in order not to reuse already created context instead of creating new one excessively.
 - Remove hardcoded fallback versions across frameworks and containers. Eight components (`new_relic, open_telemetry_javaagent, postgresql_jdbc, app_dynamics, jacoco_agent, spring_auto_reconfiguration, groovy,
    spring_boot_cli`) silently fell back to a hardcoded version string when `DefaultVersion` failed a manifest lookup. This masked misconfigured or missing manifest entries, potentially installing a stale cached artifact or failing with a confusing downstream error. All eight now return a descriptive error immediately, consistent with the rest of the codebase.
  - Fix broken classpath glob in JavaMainContainer.buildClasspath() (java_main.go). The glob pattern filepath.Join(buildDir, "$HOME/*.jar") produced a literal path like /tmp/build/$HOME/*.jar which never matched anything, silently dropping all root-level JARs from the classpath. Fixed to filepath.Join(buildDir, "*.jar") with $HOME/ prepended to each match when constructing the runtime path, consistent with how other classpath entries are built.

### DIFF
--- a/src/java/containers/groovy.go
+++ b/src/java/containers/groovy.go
@@ -47,10 +47,7 @@ func (g *GroovyContainer) Supply() error {
 	// Install Groovy runtime
 	dep, err := g.context.Manifest.DefaultVersion("groovy")
 	if err != nil {
-		g.context.Log.Warning("Unable to determine default Groovy version")
-		// Fallback version
-		dep.Name = "groovy"
-		dep.Version = "4.0.0"
+		return fmt.Errorf("unable to determine Groovy version: %w", err)
 	}
 
 	// Install Groovy with strip components to remove the top-level directory
@@ -163,3 +160,4 @@ func (g *GroovyContainer) buildClasspath() string {
 	// Adding also container security provider and the additional CLASSPATH env built when profile.d scripts are sourced
 	return "-cp " + strings.Join(jarPaths, ":") + "${CLASSPATH:+:$CLASSPATH}${CONTAINER_SECURITY_PROVIDER:+:$CONTAINER_SECURITY_PROVIDER}"
 }
+
--- a/src/java/containers/java_main.go
+++ b/src/java/containers/java_main.go
@@ -265,10 +265,8 @@ func (j *JavaMainContainer) buildClasspath() (string, error) {
 
 	// Add all JARs in the build directory
 	jarFiles, err := filepath.Glob(filepath.Join(buildDir, "*.jar"))
-	if err == nil {
-		for _, jar := range jarFiles {
-			classpathEntries = append(classpathEntries, "$HOME/"+filepath.Base(jar))
-		}
+	if err == nil && len(jarFiles) > 0 {
+		classpathEntries = append(classpathEntries, "$HOME/*.jar")
 	}
 
 	// Add lib directory if it exists
--- a/src/java/containers/java_main.go
+++ b/src/java/containers/java_main.go
@@ -264,10 +264,10 @@ func (j *JavaMainContainer) buildClasspath() (string, error) {
 	}
 
 	// Add all JARs in the build directory
-	jarFiles, err := filepath.Glob(filepath.Join(buildDir, "$HOME/*.jar"))
+	jarFiles, err := filepath.Glob(filepath.Join(buildDir, "*.jar"))
 	if err == nil {
 		for _, jar := range jarFiles {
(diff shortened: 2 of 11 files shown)


ramonskie:
makes sense..

stokpop:
Looks good, got some suggestions from my ai agent:

## Review suggestions

### 1. Classpath glob: use wildcard pattern instead of enumerating

The fix is correct, but the per-jar enumeration is inconsistent with how `lib/` and `BOOT-INF/lib/` are handled (both use `*` wildcards). Simpler approach:

```go
jarFiles, err := filepath.Glob(filepath.Join(buildDir, "*.jar"))
if err == nil && len(jarFiles) > 0 {
    classpathEntries = append(classpathEntries, "$HOME/*.jar")
}
```

Same runtime result, one classpath entry instead of N, consistent with `$HOME/lib/*` and `$HOME/BOOT-INF/lib/*` patterns already used in this function.

### 2. Add a unit test that catches the original bug

The existing `buildClasspath` tests in `java_main_test.go` don't assert the `$HOME/` prefix on root-level jars. A test like this would have caught the bug and will prevent regressions:

```go
Context("with JARs in root directory", func() {
    BeforeEach(func() {
        os.WriteFile(filepath.Join(buildDir, "app.jar"), []byte("fake"), 0644)
        os.WriteFile(filepath.Join(buildDir, "Main.class"), []byte("fake"), 0644)
    })

    It("includes root JARs with $HOME prefix in classpath", func() {
        os.Setenv("JAVA_MAIN_CLASS", "com.example.Main")
        defer os.Unsetenv("JAVA_MAIN_CLASS")

        cmd, err := container.Release()
        Expect(err).NotTo(HaveOccurred())
        // Verify root jars are reachable at runtime via $HOME path
        Expect(cmd).To(ContainSubstring("$HOME/*.jar"))
        // Or if enumerating individually:
        // Expect(cmd).To(ContainSubstring("$HOME/app.jar"))
    })
})
```

This test fails on `main` (the `$HOME/*.jar` literal path in `filepath.Glob` never matches) and passes with the fix.

stokpop (approved):

kiril-keranov:
> Looks good, got some suggestions from my ai agent:
> 
> ## Review suggestions
> ### 1. Classpath glob: use wildcard pattern instead of enumerating
> The fix is correct, but the per-jar enumeration is inconsistent with how `lib/` and `BOOT-INF/lib/` are handled (both use `*` wildcards). Simpler approach:
> 
> ```go
> jarFiles, err := filepath.Glob(filepath.Join(buildDir, "*.jar"))
> if err == nil && len(jarFiles) > 0 {
>     classpathEntries = append(classpathEntries, "$HOME/*.jar")
> }
> ```
> 
> Same runtime result, one classpath entry instead of N, consistent with `$HOME/lib/*` and `$HOME/BOOT-INF/lib/*` patterns already used in this function.
> 
> ### 2. Add a unit test that catches the original bug
> The existing `buildClasspath` tests in `java_main_test.go` don't assert the `$HOME/` prefix on root-level jars. A test like this would have caught the bug and will prevent regressions:
> 
> ```go
> Context("with JARs in root directory", func() {
>     BeforeEach(func() {
>         os.WriteFile(filepath.Join(buildDir, "app.jar"), []byte("fake"), 0644)
>         os.WriteFile(filepath.Join(buildDir, "Main.class"), []byte("fake"), 0644)
>     })
> 
>     It("includes root JARs with $HOME prefix in classpath", func() {
>         os.Setenv("JAVA_MAIN_CLASS", "com.example.Main")
>         defer os.Unsetenv("JAVA_MAIN_CLASS")
> 
>         cmd, err := container.Release()
>         Expect(err).NotTo(HaveOccurred())
>         // Verify root jars are reachable at runtime via $HOME path
>         Expect(cmd).To(ContainSubstring("$HOME/*.jar"))
>         // Or if enumerating individually:
>         // Expect(cmd).To(ContainSubstring("$HOME/app.jar"))
>     })
> })
> ```
> 
> This test fails on `main` (the `$HOME/*.jar` literal path in `filepath.Glob` never matches) and passes with the fix.

@stokpop Thanks for the review, addressed the comments

stokpop (approved):
LGTM

PR #1402, opened 2026-08-25
Remove spring-boot-cli outdated container

kiril-keranov:
Summary

- Spring Boot CLI 2.7.18 is very outdated and subject of multiple critical security vulnerabilities.
image

On the other hand latest Spring Boot CLI 4.x is very different, Spring Boot CLI 4.x **removed the** `spring run` **command entirely**, making the Spring Boot CLI container non-functional (apps would fail to start with connection refused if staged with it) so it cannot be updated. Spring Boot CLI 4.x is now focused mainly on project bootstrap

- Removed the Spring Boot CLI container and all associated code, tests, fixtures, and manifest entries since there is no viable path forward with Spring Boot CLI 4

Changes

- Deleted `src/java/containers/spring_boot_cli.go` and its unit test
- Deleted `src/integration/spring_boot_cli_test.go` and all 5 fixture directories under testdata/containers/spring_boot_cli_*
- Removed `NewSpringBootCLIContainer` registration from container.go
- Removed `IsBeans/beansPattern` from groovy_utils.go and its test (was only used by Spring Boot CLI detection)
- Removed `spring-boot-cli` from manifest.yml: default_versions, url_to_dependency_map, and dependencies entries


stokpop (approved):
Clean removal, LGTM.

Minor concern: Groovy apps that previously matched Spring Boot CLI (POGO/beans pattern without main method or shebang) will now get "No suitable container found" since GroovyContainer requires a main method or shebang. This is arguably correct since SB CLI 4.x removed `spring run` anyway, but existing users with such apps will see a behavior change on upgrade. Might be worth a note in release notes.

kiril-keranov:
> Clean removal, LGTM.
> 
> Minor concern: Groovy apps that previously matched Spring Boot CLI (POGO/beans pattern without main method or shebang) will now get "No suitable container found" since GroovyContainer requires a main method or shebang. This is arguably correct since SB CLI 4.x removed `spring run` anyway, but existing users with such apps will see a behavior change on upgrade. Might be worth a note in release notes.

@stokpop Thanks for noticing this, will try to add it somewhere in notes

